PR #17, opened 2022-07-13
Add script to run cleanup commands

RicardoVaz1:
Why:
* To run all commands to format the all app

How:
* By creating the file cleancode.sh that will run all the commands

### DIFF
--- a/cleancode.sh
+++ b/cleancode.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+echo "---------- Running ESLint ----------"
+npx eslint --fix frontend/.
+echo "------------------------------------
+"
+
+echo "--------- Running Prettier ---------"
+npx prettier --write frontend/.
+echo "------------------------------------
+"
+
+echo "---------- Running Rubocop ---------"
+rubocop -A 
+echo "------------------------------------"


cleancode.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+echo "---------- Running ESLint ----------"
+npx eslint --fix frontend/.
+echo "------------------------------------
+"
```
joaojusto, line 6: Any reason for the end quote `"` to be here instead on line 6?

cleancode.sh
```diff
@@ -0,0 +1,15 @@
...
+
+echo "--------- Running Prettier ---------"
+npx prettier --write frontend/.
+echo "------------------------------------
+"
```
joaojusto, line 11: Same question as above

joaojusto (changes requested):

cleancode.sh
RicardoVaz1: It's one way of printing a " \n " in bash. We can find a better way if it's not the best way to do it

RicardoVaz1: We can also use `echo -e "------------------------------------\n\n"`, which one should we use it?

mirhamasala: I'd say the latter is more clear. This way it doesn't look like a mistake. ☺️

joaojusto (approved):
LGTM 🚢

cleancode.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
DavideSilva, line 1: I recommend passing the `-u` and `-e` flags to bash

Instead of `#!/bin/bash`, it should be `#!/bin/bash -ue`. In short, the `-u` flag "tells the shell to treat expanding an unset parameter an error, which helps to catch e.g. typos in variable names", while the `-e`flag "tells the shell to exit if a command exits with an error"

If you want to read more why [you can read this](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) or [this](https://unix.stackexchange.com/questions/595249/what-does-the-eu-mean-in-bin-bash-eu-at-the-top-of-a-bash-script-or-a) if you are having trouble sleeping 😂

mirhamasala (dismissed):

DavideSilva:
We have a convention in the Universe to name this type of script `lint` and storing it in the `bin` folder instead of in the root directory

cleancode.sh
DavideSilva: One way of adding a new line is by simply calling the `echo` command without any arguments

PR #19, opened 2022-07-14
Jr/fixing eslint errors

jreis13:
Why:
* ESLint was throwing a number of errors related to variable naming, using let over const, react imports etc.

How:
* Replaced let with const in variables that were not being re-defined, replacing snake_case with camelCase where appropriate, added eslint rules to not require react imports since we don't need to do so in this react version

### DIFF
--- a/frontend/src/pages/Articles.js
+++ b/frontend/src/pages/Articles.js
@@ -1,67 +1,43 @@
-import React, { useState } from 'react'
-import axios from "axios"
+import React from "react";
 
-import "./Article.css"
+import "./Article.css";
 
 const Articles = () => {
-    /*let [articles, setArticles] = useState([])
-    let port = 3000 // backend port
+  const articles = [
+    { id: 1, title: "test1", autor: "test2", content: "test3" },
+    { id: 2, title: "test4", autor: "test5", content: "test6" },
+  ];
 
-    axios
-        .get("http://localhost:" + port + "/articles/")
-        .then((res) => {
-            setArticles({ articles: res.data });
-        })
-        .catch((error) => console.log(error));*/
+  function handleClick(ID) {
+    console.log("ID: ", ID);
+  }
 
-    let articles = [
-        { id: 1, title: "test1", autor: "test2", content: "test3" },
-        { id: 2, title: "test4", autor: "test5", content: "test6" }
-    ]
+  return (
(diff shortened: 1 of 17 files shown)


frontend/.eslintignore
```diff
@@ -0,0 +1,9 @@
+# Ignore 
+
+node_modules/
+frontend/node_modules/
+.*
+Gemfile 
+config.ru 
+Rakefile 
+README.md
```
mirhamasala, line 9: This is not only "Eslint" related. You're also ignoring other files. Plus, why ignore changes to README?

frontend/src/App.js
```diff
@@ -1,3 +1,4 @@
+import React from "react";
```
mirhamasala, line 1: This shouldn't be necessary in REact 17+ 

https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint

frontend/src/pages/Articles.js
```diff
@@ -1,67 +1,45 @@
...
+  const articles = [
+    { id: 1, title: "test1", autor: "test2", content: "test3" },
+    { id: 2, title: "test4", autor: "test5", content: "test6" },
+  ];
+
+  /* eslint-disable */
+  function handleClick(ID) {
+    console.log("ID: ", ID);
+  }
+  /* eslint-enable */
```
mirhamasala, line 15: Why the disable?

mirhamasala (changes requested):

frontend/src/pages/Articles.js
mirhamasala: Sorry, I just read the commit message. As discussed yesterday, please don't leave any `console.log` in your code. Also not for "testing" purposes. There should be no "fake" code that you're pushing up.

mirhamasala (commented):
Please note that this commit is called `Added eslint rules to not require the import react`,  but it also does other things, like remove commented code, etc.

Make sure your commit only does one thing. (No need to change this time, just keep it in mind.)

mirhamasala:
@jreis13 Make sure to rewrite your commit message. I think the `how` has changed.

mirhamasala (approved):
Yay, crew! LGTM! ❤️

PR #20, opened 2022-07-14
Fixing the script that runs the linters and updating README.md

beasrodrigues24:
Why:
* The paths it contained allowed to run it from the bin/ folder, which is not the standard way. It should work from the main directory of the project.
* Now eslint is run with npm run eslint, so it uses another script.
How:
* Changed the paths to match the main directory
* Changed the permissions of the script to be runnable with ./bin/lint
* Updated this information on README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Subvisuwheel
 
-App that runs a game where the user is allowed to spin a wheel for prizes. 
+App that runs a game where the user is allowed to spin a wheel for prizes.
 
 The backend is built with Ruby on Rails while the frontend is built with React.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -15,12 +15,18 @@ The backend is built with Ruby on Rails while the frontend is built with React.
...
-``npm install``
-``npm start``
+
+`cd frontend/`
+`npm install`
+`npm start`
+
+### Eslint, Prettier and Rubocop 
+
+`./bin/lint`
```
mirhamasala, line 32: What is this line supposed to tell the developer trying to set up your project? What do you want them to know? Where the script is located? How to run the script?

mirhamasala (approved):

PR #21, opened 2022-07-19
Connecting the back-end and front-end of the wheel

jreis13:
Why:
* Solving the core of the backend for the wheel, back-end and front-end connection.

How:
* Editing the controller so it shows the data as we want it, pair-working with the mentors that helped us implement this feature.


db/migrate/20220711110429_create_articles.rb
```diff
@@ -1,11 +0,0 @@
...
-  def change
-    create_table :articles do |t|
-      t.string :author
-      t.string :title
-      t.string :content
-
-      t.timestamps
-    end
-  end
-end
```
zediogoviana, line 11: Be careful when doing this... As I explained in the first day when we setup Rails, changing a migration file does nothing unless it's rollbacked (and that should happen during the development phase or a major issue when deploying). This way, deleting a migration file does nothing -> you'll have the Articles table forever in your Database unless you Drop the full database and start over again (You can already see why this is a problem and should be avoided, but as you haven't anything deployed, we can assume all of you will start fresh again).

But again, be careful with this in the future.

db/migrate/20220718131723_create_wheel_configurations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateWheelConfigurations < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wheel_configurations do |t|
+      t.references :wheel
+      t.string :color_1
+      t.string :color_2
```
zediogoviana, line 6: Why not call this something more explicit like `main_color` and `secondary_color`? Just a suggestion comment, don't need to update if you all agree with these namings.

db/migrate/20220718133933_create_prizes.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePrizes < ActiveRecord::Migration[7.0]
+  def change
+    create_table :prizes do |t|
```
zediogoviana, line 3: Since you have the `presence: true` constraint in the Model definition, why not add it here also? Don't trust that much in the framework, and this way you'll also remember t do it in other languages that don't give those constraints for free.

db/migrate/20220718133957_create_wheels.rb
```diff
@@ -0,0 +1,10 @@
+class CreateWheels < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wheels do |t|
+      t.string :name
```
zediogoviana, line 4: Same as above

config/routes.rb
```diff
@@ -1,4 +1,8 @@
 Rails.application.routes.draw do
-  get "/articles", to: "articles#index"
-  get "/articles/:id", to: "articles#show"
+  # ----- Wheel Configuration ----- #
+  # get "/wheel", to: "wheels#index"
+  get "/prizes", to: "prizes#index"
+  get "/wheel-configurations", to: "wheel_configurations#index"
```
zediogoviana, line 5: If you don't need the `/wheel` endpoint you should remove it entirely.
Also, about the `# ----- Wheel Configuration ----- #`, I think this seems a bit redundant as this should be enough to understand what's happening in this case. Take a look [here](https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing), maybe you could do something like this:

```
scope '/wheel' do
  get "/prizes", ...
  get "/configuration", ...
end
```

Looking at it you wouldn't need the comment above as everything would be under the `wheel` scope.

app/controllers/wheel_configurations_controller.rb
```diff
@@ -0,0 +1,17 @@
+class WheelConfigurationsController < ApplicationController
+    def index
+        render json: WheelConfiguration.all
+    end
+
+    def create
+        wheel_configuration = WheelConfiguration.create(wheel_configuration_params)
+        render json: wheel_configuration
+    end
+
```
zediogoviana, line 10: Double blank line, you just need one 😅

zediogoviana (changes requested):
I think this PR is too big and does a lot of things. Things that are not described in the description, and things that could be in separate PRs. For example, removing the articles part should be a completely different PR.

Maybe we could leave that, this time, but 32 files is a lot to review unless really necessary.

jreis13:
> 

We split the commits into different PR's and fixed this issue.

db/migrate/20220711110429_create_articles.rb
jreis13: Understood.

db/schema.rb
```diff
@@ -22,4 +22,30 @@
...
+  end
+
+  create_table "wheel_configurations", force: :cascade do |t|
+    t.bigint "wheel_id"
+    t.string "color_1"
+    t.string "color_2"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["wheel_id"], name: "index_wheel_configurations_on_wheel_id"
+  end
```
zediogoviana, line 42: You need to run the migrations to have the schema updated. It's still showing here with `color_1` instead of the new naming

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
-  get "/prizes", to: "prizes#index"
-  get "/configurations", to: "wheel_configurations#index"
+  get '/wheel', to: 'wheel#show'
```
zediogoviana, line 2: This won't work until you actually change the `show` method to always return the same thing. The show expects an `id` to be passed, and this case you are not passing anything.

app/controllers/wheels/prizes_controller.rb
```diff
@@ -0,0 +1,16 @@
+class PrizesController < ApplicationController
```
zediogoviana, line 1: Have a look at how to namespace controllers, as I suggested. This and the `configurations` one should have a different naming, now that they're namespaced.

app/controllers/wheel_configurations_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    private
+
+    def wheel_configuration_params
+        params.require(:wheel_configuration).permit(:color_1, :color_2)
```
zediogoviana, line 15: Have you been testing the requests through Postman or anything else? This won't work until you update all references of `color_1` and `color_2`.

Testing things before committing it's very important.

frontend/src/components/Wheel/index.js
```diff
@@ -1,89 +1,72 @@
...
+          <th>Name</th>
+          <th>Logo</th>
+          <th>Prizes</th>
+          <th>Color1</th>
+          <th>Color2</th>
+        </tr>
 
-  const prizes = ["Prize1", "Prize2", "Prize3", "Prize4"];
+        {/* {wheelInfo.map((wheel) => {
+          return ( */}
```
zediogoviana, line 39: Please remove the comments in this file, unless they're really necessary

frontend/src/components/Wheel/index.js
```diff
@@ -1,89 +1,72 @@
...
-        { percentage: "0.30", color: "yellow" },
-      ]}
-    />
-  );
-};
-
-export const PieCircle = (props) => {
-  let totalPercentage = 0;
+const Wheel = () => {
+  const [wheelInfo, setWheelInfo] = useState([]);
```
zediogoviana, line 6: The default here is not a list anymore I believe, as you have changed that in the backend.

frontend/src/components/Wheel/index.js
```diff
@@ -1,89 +1,72 @@
...
+        if (response.ok) {
+          return response.json();
+        } else {
+          throw response;
+        }
+      })
+      .then((data) => {
+        setWheelInfo(data)
+        setMainColor(data.config.main_color)
+        setSecondaryColor(data.config.secondary_color)
```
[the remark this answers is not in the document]
RicardoVaz1: Done

frontend/src/components/Wheel/index.js
RicardoVaz1: Done

frontend/src/components/Wheel/index.js
RicardoVaz1: Done

zediogoviana (approved):
👍

PR #24, opened 2022-07-20
Experiment the look of the Wheel

RicardoVaz1:
Why:
* It will be easier to do the Wheel w/ right design after experimenting

How:
* Tutorial: https://www.youtube.com/watch?v=2plnsl8P2zE
* Personalize the Wheel to load the prizes on the Backend
* Allow customizeble number of prizes


mirhamasala:
@RicardoVaz1 Let's park the review for this until you make the changes discussed today!

mirhamasala:
FYI You can convert PRs back to a draft until they're ready to be reviewed. In this case, we have some other features and fixes pending before we can get to this.

Screenshot 2022-07-21 at 18 06 11

RicardoVaz1:
@mirhamasala I fixed the merge conflicts

mirhamasala:
> @mirhamasala I fixed the merge conflicts

You still need to rebase though. I'm seeing commits in the PR that have already been merged into `main`.

RicardoVaz1:
We still need to fix this code before the Code Review. We need to change the load of the prizes

frontend/src/components/Wheel/index.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stop-rotate {
+  animation-play-state: paused;
+}
+
+.img_prizes {
+  width: 30px;
+  height: 30px;
+}
```
joaojusto, line 30: We should avoid inconsistent naming conventions: `img_prizes` vs `stop-rotate` for example.

frontend/src/components/Wheel/index.css
```diff
@@ -0,0 +1,30 @@
+.circle {
+  width: 25em;
+  height: 25em;
```
joaojusto, line 3: Why are we using `em`s ? And why are we using them here if in other places we use `px`s?

frontend/src/components/Wheel/index.css
```diff
@@ -0,0 +1,30 @@
...
+  border: 1px solid black;
+  position: relative;
```
joaojusto, line 5: I don't understand why do we need position `relative` here, can you elaborate?

frontend/src/components/Wheel/index.js
```diff
@@ -1,12 +1,39 @@
 import { useEffect, useState } from "react";
-import SpinButton from "./button";
 import styles from "./index.module.css";
+import "./index.css";
 
 const Wheel = () => {
   const [wheelInfo, setWheelInfo] = useState({});
+  const [rotate, setRotate] = useState("circle");
```
joaojusto, line 7: I don't think this is a good idea, it's error prone, you need to know change the whole class name when toggling states and it hiddes what is doing (indirection).

I would suggest using a bool as state, `isRotating` and then using the package `classnames` to toggle the correct classes.

frontend/src/components/Wheel/index.js
```diff
@@ -1,12 +1,39 @@
...
+
+  async function startRotate() {
```
joaojusto, line 9: Why is this function `async`?

frontend/src/components/Wheel/index.js
```diff
@@ -19,46 +46,28 @@ const Wheel = () => {
...
-            wheelInfo?.config?.secondaryColor,
-          ]}
-        />
-      </div>
+      <h1 className={styles.center}>{wheelInfo.name}</h1>
+      <div className={styles.arrow}></div>
+      <ul className={rotate}>
+        {wheelInfo?.prizes?.map((prize, i) => {
+          return (
+            <li className={styles.prize} key={prize.id} style={wheelStyle(i)}>
```
joaojusto, line 54: This isn't a list if every list only has one item inside. Do you mean moving `<li>` before `wheelInfo.prizes.map`?

frontend/src/components/Wheel/index.module.css
```diff
@@ -3,3 +3,84 @@
...
+}
+
+.stop-rotate {
+  animation-play-state: paused;
+}
+
+.img_prizes {
+  width: 30px;
+  height: 30px;
+}
```
joaojusto, line 86: Why do we have two stylesheets for this component? Which one is used?

joaojusto (changes requested):

frontend/src/components/Wheel/index.js
```diff
@@ -1,65 +1,82 @@
 import { useEffect, useState } from "react";
-import SpinButton from "./button";
 import styles from "./index.module.css";
+import "./index.css";
```
mirhamasala, line 3: Why?

frontend/src/components/Wheel/index.js
```diff
@@ -1,65 +1,82 @@
...
-                <div key={prize.id}>
-                  <span>Name: {prize.name}</span>
-                  <span> Percentage: {prize.percentage}</span>
-                  <span> Image: {prize.image}</span>
-                </div>
-              );
-            })}
-          </td>
+    <>
+      <h1 className={styles.center}>Prizes</h1>
```
mirhamasala, line 52: 🙄

frontend/src/components/Wheel/index.js
```diff
@@ -1,65 +1,82 @@
...
+              className={styles.prize}
+              key={prize?.id}
+              style={wheelStyle(index)}
+            >
+              <div contentEditable="true" spellCheck="false">
+                <span className={styles.text}>{prize?.name}</span>
+                <img
+                  src={prize?.image}
+                  alt={prize?.name}
+                  className={styles.img_prizes}
```
mirhamasala, line 68: We're using CSS modules: `imgPrizes` Though I don't understand what image prizes are? If it's only one image?

frontend/src/components/Wheel/index.js
```diff
@@ -1,65 +1,82 @@
...
-      <div className={styles.align}>
-        <SpinButton
-          colors={[
-            wheelInfo?.config?.main_color,
-            wheelInfo?.config?.secondaryColor,
-          ]}
-        />
-      </div>
-    </div>
+      <button className={styles.spin_button} onClick={startRotate}>
```
mirhamasala, line 76: Same as above

frontend/src/pages/WheelPage/index.js
```diff
@@ -1,3 +1,10 @@
-const WheelPage = () => <div>This is the wheel page.</div>;
+import Wheel from "../../components/Wheel";
```
mirhamasala, line 1: Use absolute imports

app/controllers/prizes_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    render json: Prize.all
+  end
+
+  def create
+    prize = Prize.create(prize_params)
+    render json: prize
+  end
+
+  def win_prize
+    render json: Prize.where(image: 'Logo  2')
```
mirhamasala, line 12: `Logo  2` seems to have to spaces. Also, can we please be more descriptive in naming? What's logo 2?

beasrodrigues24: That's a test I forgot to remove, I'll make sure to remove that tomorrow. That should be some logic to give a prize at random according to the probabilities.

frontend/src/components/Wheel/index.js
beasrodrigues24: We changed that to prizesInfo.map((prize, index), and that's a list with multiple objects (prizes). Aren't we supposed to have each one of them being an item?

frontend/src/components/Wheel/index.module.css
```diff
@@ -1,5 +1,127 @@
...
+
+@media only screen and (max-width: 350px) {
+  .wheelName {
+    margin-left: 65px !important;
+  }
+
+  .arrow {
+    margin-right: -60px !important;
+  }
```
DavideSilva, line 87: We should probably remove this `important` at same stage in the future

frontend/src/components/Wheel/index.js
beasrodrigues24: By this time we didn't know why it should be async and we removed it. Then we put a fetch to the back-end to give the prize the user as won and determined that the fetch should have an await (to not stop spinning before a prize is awarded) and therefore the startRotate actually ended up async again.

frontend/src/components/Wheel/index.module.css
DavideSilva: Stylesheet was removed

DavideSilva (approved):
In the future, we should have smaller PR's, like we repeatedly have mentioned, but, in the sense of moving things along, I'm going to approve and let them merge this. Otherwise, we will spend the next week dealing with PR's and rebases and what not...

:shipit:

DavideSilva:
@joaojusto @mirhamasala I'm going to merge this. I was with the campers in the office and I did a review and we agreed to let this one slide so that we can move forward on Monday with a clean slate

frontend/src/components/Wheel/index.module.css
mirhamasala: Why are we merging PRs that aren't right al ready?

PR #25, opened 2022-07-21
Delete unused left-over files

jreis13:
Why:
* Files were leftover from testing phase;
* Most files were empty and / or not being used.

How:
* Deleted useless files.

### DIFF
--- a/frontend/src/components/CheckBox/index.js
+++ b/frontend/src/components/CheckBox/index.js
@@ -1,7 +1,0 @@
-import "./index.css";
-
-const CheckBox = (props) => {
-  return <div className="">CheckBox</div>;
-};
-
-export default CheckBox;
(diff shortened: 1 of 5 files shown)


mirhamasala (changes requested):
`Signup.js` is complaining on my side. It seems to be using `CheckBox`, `DropdownBox` and `InputText`.

Did you run the project after deleting these components?

jreis13:
> `Signup.js` is complaining on my side. It seems to be using `CheckBox`, `DropdownBox` and `InputText`.
> 
> Did you run the project after deleting these components?

I noticed this after and I believe its fixed on the next PR i made

mirhamasala:
> > `Signup.js` is complaining on my side. It seems to be using `CheckBox`, `DropdownBox` and `InputText`.
> > Did you run the project after deleting these components?
> 
> I noticed this after and I believe its fixed on the next branch/PR i made

It should be on this PR, though. If we merge this PR as is now, the app would break.

mirhamasala (approved):

PR #27, opened 2022-07-21
Fixing props problems

beasrodrigues24:
Why:
* Props.content was unnecessary since children is a prop more appropriate to pass information like text
* We were being redundant in the way we were accessing the props (props.this, props.that)
* Header component didn't use its props

How:
* Replacing in components the use of <Component content="x"> by <Component>x</Component> and accessing it as {children} in the component index.js
* Replacing const Component = (props) => ...  by const Component = ({this,that}) => ...
* Placing the use of children in the header component


mirhamasala:
@beasrodrigues24 Make sure to rebase, fix conflicts, and push your branch up again - Otherwise it can't really be reviewed. 🙏🏼

beasrodrigues24:
@mirhamasala  Done, i fixed it. Don't think i used the right commands this time but will next time :)

mirhamasala:
> @mirhamasala Done, i fixed it. Don't think i used the right commands this time but will next time :)

I think you should try again. Delete the last commit you made. Then, pull the remote `main` into your local `main` and go back to this branch to rebase. Fix conflicts. Then push up again. If you need help, let me know. 🙏🏼

frontend/src/components/Button/index.js
```diff
@@ -1,12 +1,13 @@
 import { PropTypes } from "prop-types";
 import styles from "./index.module.css";
 
-const Button = (props) => {
-  return <button className={styles.button}>{props.children}</button>;
-};
+const Button = ({children, onClick}) => {
+  return <button className={styles.button} onClick={onClick}>{children}</button>;
```
mirhamasala, line 5: You don't have an `onClick` function defined.

mirhamasala: You can remove it for now.

mirhamasala (changes requested):
Except for the one comment and fixing your Git history, everything else looks good.

frontend/src/components/Button/index.js
mirhamasala: I'm surprised your linter didn't complain? Can you check if it's set up correctly before you make the change?

beasrodrigues24: Okay I'll remove it. I put it because since you said onClick was a prop I thought I could pass a handler inside this button and it would be useful in the future :sweat_smile:

frontend/src/components/Button/index.js
```diff
@@ -1,12 +1,13 @@
...
+}
 
 Button.propTypes = {
   children: PropTypes.string.isRequired,
+  onClick: PropTypes.string.isRequired
```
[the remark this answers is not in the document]
beasrodrigues24: During the merge I messed it up, before that I had PropTypes.onClick , but I wasn't sure if that was a thing either :)

frontend/src/pages/Home.js
```diff
@@ -0,0 +1,33 @@
+// Components
```
mirhamasala, line 1: This page shouldn't exist anymore. You have `HomePage` now. Probably leftover from rebase.

frontend/src/pages/Signup.js
```diff
@@ -0,0 +1,39 @@
+import "./signup.css";
```
mirhamasala, line 1: Same. This page doesn't exist anymore.

frontend/src/components/Button/index.js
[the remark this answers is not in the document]
mirhamasala: Next time, you may want to check the [documentation](https://github.com/facebook/prop-types) to find out. 🙂

frontend/src/components/Button/index.js
mirhamasala: It's a prop and we'll need it but at the moment you don't have any functionality. There's no spec on what to do `onClick.` We'll get to it once we start properly using the `Button` component. Also, some other attributes are missing, so we're going to need to check the Lint settings, as Justo mentioned.

frontend/src/pages/Home.js
beasrodrigues24: I don't know how this happened :) 
So how do I proceed to remove this? Rebase again?

mirhamasala: I'm guessing you may've chosen the "wrong thing to keep" when solving conflicts.

mirhamasala: I think rebasing again won't work.

You can delete the pages and do `git commit --amend --no-edit` which will add those changes to your last commit. Then, I'd make sure to squash your 2 commits into one. But Git master @DavideSilva may have another suggestion.

beasrodrigues24: I think it's done now :)

DavideSilva: No, that's usually what I do when I mess up and end up with changes that I don't want 👍

jreis13 (approved):

PR #33, opened 2022-07-22
added slug file

jreis13:
Why:
* We want to find a way to not upload the frontend folder to the Heroku server as we will be running the frontend in another server.

How:
* Added the frontend folder to the `.slugignore` file.

### DIFF
--- a/.slugignore
+++ b/.slugignore
@@ -1,0 +1,1 @@
+frontend/


DavideSilva (approved):
:shipit:

DavideSilva:
Just a small note. PR titles and commit messages are usually in the present tense. 

"This commit _does_ this", instead of "this commit _did_ something"

PR #37, opened 2022-07-26
Fix lint config

mirhamasala:
Why:

- Rules and config was being overridden

How:

- Remove unnecessary packages > Create React App already uses `import` and `jsx-a11y`
- Clean up custom rules that are already being applied via ESLint in `react-app`
- Add `jsx-a11y/recommended` for more accessibility rules
- Change `no-console` rules to `warn` because having it set to `error` is way annoying
- Run ESLint and fix problems


joaojusto (approved):
LGTM

PR #38, opened 2022-07-26
Add Constants and Fix SVGs

RicardoVaz1:
Why:
* To fix hard code
* SVG's didn't look good and were inconsistent

How:
* Adding a constants file
* Calling the variables from the constants file
* Replaced them across the code

Co-authored-by: Beatriz Rodrigues <beasrodrigues24@users.noreply.github.com>


frontend/src/pages/PrizePage/index.module.css
```diff
@@ -1,3 +1,10 @@
-.center {
+.root {
   text-align: center;
 }
+
+.item {
+  margin: 20px 0px;
+  width: 120px;
+  height: 120px;
+  margin: auto;
```
mirhamasala, line 9: You're overwriting line 6 here.

frontend/src/pages/PrizePage/index.js
```diff
@@ -1,25 +1,44 @@
...
+    prize === "nft" ? setPrize("tshirt") : setPrize("nft");
   };
 
   return (
-    <div className={styles.center}>
+    <div className={styles.root}>
       <button onClick={handleClick}>Change Prize</button>
-      <Prize name={prize} />
+
+      <div className="center">
```
joaojusto, line 29: We should avoid naming classes after styles.

mirhamasala (approved):
Wow. 😭

frontend/src/pages/HomePage/index.js
```diff
@@ -8,26 +10,28 @@ import styles from "./index.module.css";
...
-        <div className={styles.container}>
-          <Title>Welcome to Subvisual&apos;s Wheel of Fortune!</Title>
+    <>
+      <Header>{COMPANY}</Header>
+      <div className={styles.title}>
+        <Title>Welcome to {COMPANY}&apos;s Wheel of Fortune!</Title>
+      </div>
+
+      <div className={styles.container}>
+        <div className={styles.inner}>
```
joaojusto, line 20: I don't think `inner` is very descriptive, what do you think about something like `callToAction`, `about` or `informationBlock`?

frontend/src/pages/SocialPage/index.js
```diff
@@ -1,33 +1,41 @@
...
+          <a href={WEBSITE} target="_blank" rel="noreferrer">
+            {WEBSITE.substring(8)}
+          </a>{" "}
+          and on our social media:
+        </Text>
       </div>
 
-      <div className={styles.item}></div>
-    </div>
+      <ul className={styles.items}>
```
joaojusto, line 26: What about `links` instead of `items`? And `link` instead of `item`?

frontend/src/pages/PrizePage/index.js
```diff
@@ -1,25 +1,44 @@
...
+        <Title>{`Congrats, you won a ${prize}`}</Title>
+
+        <div className={styles.item}>{prizes[prize]}</div>
```
joaojusto, line 32: `item` seems very generic, why not calling it `prize`?

DavideSilva (changes requested):
Why are you committing `package-lock.json` changes without any change to the `package.json`?

Looking at the file, there's new changes that I'm not sure make sense. There's a `packages` map that have an empty string as the key

joaojusto (approved):
This LGTM, I would work a little bit more on naming classes but it's up to you.

DavideSilva (approved):
Disregard my last comment, turns out it's the new `npm` default

frontend/src/pages/PrizePage/index.js
beasrodrigues24: We'll fix all of these before merging to main tomorrow morning :smile:

frontend/src/pages/HomePage/index.js
```diff
@@ -17,7 +17,7 @@ const HomePage = () => {
       </div>
 
       <div className={styles.container}>
-        <div className={styles.inner}>
+        <div className={styles.informationBlock}>
```
mirhamasala, line 20: Not sure @joaojusto would've approved `informationBlock` 😝

PR #41, opened 2022-08-01
Fix front and back-end connection

beasrodrigues24:
Why:
* Cors was allowing only localhost:3001 as origin 

How:
* Added a new possible origin: subvisuwheel.netlify.app in cors.rb.
* Kept localhost:3001 for local testing purposes

### DIFF
--- a/config/initializers/cors.rb
+++ b/config/initializers/cors.rb
@@ -7,7 +7,7 @@
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    origins "localhost:3001"
+    origins "subvisuwheel.netlify.app"
 
     resource "*",
       headers: :any,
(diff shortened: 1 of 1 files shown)


config/initializers/cors.rb
```diff
@@ -7,7 +7,7 @@
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
   allow do
-    origins "localhost:3001"
+    origins "subvisuwheel.netlify.app"
```
DavideSilva, line 10: This will make it so that only the frontend app at that URL is able to call the backend app. Probably not what we want as we also need to keep `localhost:3001` for local development

Look into how to allow multiple origins for this

DavideSilva (changes requested):
Check my comment

config/initializers/cors.rb
beasrodrigues24: Okay, I will :+1:

beasrodrigues24:
I have to fix the 3000 to 3001

DavideSilva (approved):
:shipit:

PR #43, opened 2022-08-01
Save awarded prize

RicardoVaz1:
Why:
* [Save awarded prize](https://3.basecamp.com/4319812/buckets/26581088/todos/5183204497)
* To save the win prize on the Backend

How:
* Using a fetch POST w/ userEmail and prizeName

### DIFF
--- a/app/controllers/leads_controller.rb
+++ b/app/controllers/leads_controller.rb
@@ -2,6 +2,10 @@
 
 # Controller that manages leads
 class LeadsController < ApplicationController
+  def index
+    render json: Lead.all
+  end
+
   def create
     lead = Lead.create(leads_params)
     render json: lead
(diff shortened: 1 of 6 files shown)


DavideSilva (changes requested):
This has some conflicts that should be resolved. A PR should only be opened, or marked as ready to review if it were a draft, when it doesn't have any conflicts.

Also, please link the Basecamp task in the description like Justo asked on Slack

Also, let's wait for https://github.com/Subvisual-Academy/subvisuwheel/pull/58 to be merged. The PrizesController#win_prize method to will need to be updated. Instead of  updating`prizes_id` to have the prize id, we can simply call `lead.update(prize: prize)`, without having to pass ids. But this needs to wait for PR #58 first as that will change the database

RicardoVaz1:
@DavideSilva I already made the changes. Do I need to change anything else?

mirhamasala:
Please remember to fixup and rebase the commits before merging. 🙏🏼

PR #47, opened 2022-08-01
Update body text component

jreis13:


frontend/src/pages/HomePage/index.js
```diff
@@ -18,11 +18,11 @@ const HomePage = () => {
...
       <div className={styles.container}>
         <div className={styles.informationBlock}>
-          <Text>
+          <Body>
             Spin the {WHEEL_CONFIG.name} to win merch, experiences, NFTs and
             more.
-          </Text>
-          <Text>What are you waiting for? Let&apos;s get started!</Text>
+          </Body>
+          <Body>What are you waiting for? Let&apos;s get started!</Body>
```
mirhamasala, line 25: Can you explain why you chose to go with `body` 

I'm asking because there's an HTML element called `body`. Usually you'll only find one of those, so seeing multiple bodies is a bit confusing for me. But also, as a general name choice for a component, it doesn't tell me much. In fact, even if you went with `Body` I would've probably expected to see layout styles applied.

mirhamasala: I think `Text` was quite okay?

joaojusto: @mirhamasala This was to match the naming Hanna has [over here](https://www.figma.com/file/36e7nnrNlYQENyigL9sQED/Wheel-of-Fortune?node-id=496%3A2)

jreis13: Yes that was what I aimed for. That said we all agree that I should rename them `Text` again?

joaojusto (approved):
As soon as @mirhamasala is happy I'm ok shipping this 👍

frontend/src/components/Typography/Body/index.module.css
```diff
@@ -1,5 +1,5 @@
 .body {
-  font-family: var(--font-family-body);
+  font-family: var(--font-family-body), sans-serif;
```
mirhamasala, line 2: Hmm so ... would this be something to add to the global var?

frontend/src/pages/HomePage/index.js
mirhamasala: @joaojusto Want to keep it then?

frontend/src/components/Typography/Body/index.module.css
jreis13: Added on the CSS constants pr.

frontend/src/pages/HomePage/index.js
mirhamasala: Besides the naming question. Do we need to use two `<Body>` components here, though?

PR #48, opened 2022-08-01
Add Input Component

RicardoVaz1:
Why:
* To use it later for text and email inputs

How:
* Creating the Input component

### DIFF
--- a/frontend/src/components/Input/index.js
+++ b/frontend/src/components/Input/index.js
@@ -1,0 +1,36 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.css";
+import classNames from "classnames";
+
+const Input = ({ type, placeholder, disabled, error, errorMessage }) => {
+  const inputClassNames = classNames(styles.input, {
+    [styles.hasError]: error,
+  });
+
+  return (
+    <>
+      {error ? <p className={styles.errorMessage}>{errorMessage}</p> : null}
+      <input
+        disabled={disabled}
+        type={type}
+        placeholder={placeholder}
+        className={inputClassNames}
+      />
+    </>
+  );
+};
+
+Input.propTypes = {
+  type: PropTypes.oneOf(["text", "email"]).isRequired,
+  placeholder: PropTypes.string.isRequired,
+  errorMessage: PropTypes.string,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
+};
+
+Input.defaultProps = {
+  disabled: false,
+  type: "text",
+};
+
+export default Input;
--- a/frontend/src/components/Input/index.module.css
+++ b/frontend/src/components/Input/index.module.css
@@ -1,0 +1,39 @@
+.errorMessage {
+  margin-top: 0;
+  margin-bottom: 12px;
+  font-size: var(--font-size-xxs);
+  line-height: var(--font-size-xs);
+  color: var(--color-red);
+}
+
+.input {
+  padding: 20px 24px;
+  font-family: var(--font-family-body);
+  font-size: var(--font-size-xs);
+  line-height: var(--font-size-sm);
+  color: var(--color-black);
+  border: 1px solid var(--color-purple);
+  border-radius: 10px;
+  /* stylelint-disable-next-line color-function-notation */
+  /* stylelint-disable-next-line alpha-value-notation */
+  box-shadow: 0 4px 5px rgba(159, 24, 217, 0.5);
+}
+
+.input:disabled {
+  color: var(--color-dark-gray);
+  background-color: var(--color-light-gray);
+  border: 1px solid var(--color-dark-gray);
+  box-shadow: 0 4px 5px var(--color-gray);
+}
+
+.input::placeholder {
+  color: var(--color-light-gray);
+}
+
+.input:disabled::placeholder {
+  color: var(--color-dark-gray);
+}
+
+.hasError {
+  border-color: var(--color-red);
+}
--- a/frontend/src/components/Input/index.module.css
+++ b/frontend/src/components/Input/index.module.css
@@ -14,9 +14,7 @@
   color: var(--color-black);
   border: 1px solid var(--color-purple);
   border-radius: 10px;
-  /* stylelint-disable-next-line color-function-notation */
-  /* stylelint-disable-next-line alpha-value-notation */
-  box-shadow: 0 4px 5px rgba(159, 24, 217, 0.5);
+  box-shadow: 0 4px 5px rgba(159, 24, 217, 50%);
 }
 
 .input:disabled {
--- a/frontend/src/styles/globals.css
+++ b/frontend/src/styles/globals.css
@@ -13,6 +13,7 @@
   --color-gray: #808080;
   --color-light-gray: #bfbfbf;
   --color-white: #fff;
+  --color-red: #f73f52;
   --font-family-body: "Work Sans", sans-serif;
   --font-family-title: "Spicy Rice", cursive;
   --font-size-xs: 16px;
--- a/frontend/.stylelintrc
+++ b/frontend/.stylelintrc
@@ -20,6 +20,7 @@
     "selector-class-pattern": [
       "^[a-z][a-zA-Z0-9]+$",
       { "message": "Expected custom property name to be lowerCamelCase" }
-    ]
+    ],
+    "color-function-notation": "legacy"
   }
 }


frontend/src/components/Input/index.js
```diff
@@ -0,0 +1,43 @@
...
+import styles from "./index.module.css";
+
+const Input = ({ inputType, disabled, error }) => {
+  let texts = {
+    text: { placeholder: "Ariana", errorMessage: "Enter your name." },
+    email: {
+      placeholder: "arina@fortunemail.com",
+      errorMessage: "Invalid email.",
+    },
+  };
```
mirhamasala, line 11: If it's possible, I would prefer to see these used directly. I find this a bit confusing.

frontend/src/components/Input/index.js
```diff
@@ -0,0 +1,43 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.css";
+
+const Input = ({ inputType, disabled, error }) => {
```
mirhamasala, line 4: Just `type` would be fine for me.

frontend/src/components/Input/index.js
```diff
@@ -0,0 +1,43 @@
...
+            className={`${styles.input} ${styles.inputError}`}
+          />
+        </div>
+      ) : (
+        <input
+          disabled={disabled}
+          type={inputType}
+          placeholder={texts[inputType].placeholder}
+          className={styles.input}
+        />
```
mirhamasala, line 31: I would prefer to see the happy path first.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+.inputError {
+  border-color: red;
+}
+
+.container {
+  display: inline-grid;
+  font-size: 12px;
+  line-height: 16px;
+  color: #F73F52;
+}
```
mirhamasala, line 40: Let's place styles in the order they appear in the HTML. So, in this case, `container` would be a the top. Unless you go with the happy path first. Also, I'd rename this to `errorContainer` or something because when I see `container` I'm expecting a root element.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+  font-family: "Work Sans", sans-serif;
+  font-size: 16px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: 20px;
+  color: #000;
+  border: 1px solid #9f18d9;
+  border-radius: 10px;
+  box-shadow: 0 4px 5px #9f18d9 0.5;
+}
```
mirhamasala, line 12: Let's use the globals that Joao created.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+  outline-style: none;
+}
+
+.input:focus {
+  outline-style: none;
+}
+
+.input:disabled {
+  color: #404040;
+  background: #bfbfbf;
```
mirhamasala, line 25: Let's be more specific and use `background-color` instead of the shorthand `background`

frontend/src/components/Input/index.module.css
mirhamasala: I might even consider creating an `Error` component

frontend/src/components/Input/index.js
mirhamasala: Perhaps you can look into [default prop values](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) ...

I think `disabled` should default to `false`

frontend/src/components/Input/index.module.css
mirhamasala: The input styles don't look anything like the [design](https://www.figma.com/file/36e7nnrNlYQENyigL9sQED/Wheel-of-Fortune?node-id=0%3A1)?

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+  line-height: 20px;
+  color: #000;
+  border: 1px solid #9f18d9;
+  border-radius: 10px;
+  box-shadow: 0 4px 5px #9f18d9 0.5;
+}
+
+.input:active {
+  color: #000;
+  outline-style: none;
```
mirhamasala, line 16: Removing `outline` without adding back styles in a different way to signal these state changes is a really bad practice for accessibility.

Let's keep the browser defaults until we have a design for them. Or, let's not add any styles that we still don't have the design for.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+.container {
+  display: inline-grid;
+  font-size: 12px;
+  line-height: 16px;
+  color: #F73F52;
+}
+
+::placeholder {
+  color: #BFBFBF;
+}
```
mirhamasala, line 44: What do you think this does?

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+
+.inputError::placeholder {
+  color: #F73F52;
+  opacity: 1;
```
mirhamasala, line 48: The default value for `opacity` is `1`

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.input:disabled::placeholder {
+  color: #404040;
+  opacity: 1;
```
mirhamasala, line 53: Same as above.

frontend/src/components/Input/index.js
```diff
@@ -0,0 +1,43 @@
...
+      placeholder: "arina@fortunemail.com",
+      errorMessage: "Invalid email.",
+    },
+  };
+
+  return (
+    <>
+      {error ? (
+        <div className={styles.container}>
+          <span>{texts[inputType].errorMessage}</span>
```
mirhamasala, line 17: Can't we conditionally render the `error` instead of duplicating the `input`?

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+  padding: 20px 24px;
+  font-family: "Work Sans", sans-serif;
+  font-size: 16px;
+  font-style: normal;
+  font-weight: 400;
+  line-height: 20px;
+  color: #000;
+  border: 1px solid #9f18d9;
+  border-radius: 10px;
+  box-shadow: 0 4px 5px #9f18d9 0.5;
```
mirhamasala, line 11: This declaration doesn't work.

mirhamasala, line 11: This declaration doesn't work.
Screenshot 2022-08-02 at 08 48 55

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,54 @@
...
+
+.input:focus {
+  outline-style: none;
+}
+
+.input:disabled {
+  color: #404040;
+  background: #bfbfbf;
+  border: 1px solid #404040;
+  border-radius: 10px;
```
mirhamasala, line 27: You already declared `border-radius` at top. (Even though design doesn't seem to have one at all.)

frontend/src/components/Input/index.module.css
mirhamasala: This declaration doesn't work.

Screenshot 2022-08-02 at 08 51 59

frontend/src/components/Input/index.js
joaojusto: Agree, this shouldn't be here it should be configured by props

frontend/src/components/Input/index.module.css
joaojusto: @mirhamasala we are going into [hi-fi](https://www.figma.com/file/36e7nnrNlYQENyigL9sQED/Wheel-of-Fortune?node-id=496%3A2) 🥳

frontend/src/components/Input/index.module.css
joaojusto: @mirhamasala @RicardoVaz1 we need to ask Hanna what the focused/active state should look like

frontend/src/components/Input/index.module.css
mirhamasala: Awesome. My comment still holds, though. 🙈 Maybe "the styles don't look _exactly_ like the design."

frontend/src/components/Input/index.module.css
RicardoVaz1: It change the color of the placeholder text

frontend/src/components/Input/index.module.css
RicardoVaz1: Ok, I'll remove it

frontend/src/components/Input/index.module.css
RicardoVaz1: Ok, I'll remove it

frontend/src/components/Input/index.module.css
RicardoVaz1: I changed to:
```
box-shadow: 0 4px 5px #9f18d9;
opacity: 0.5;
```

frontend/src/components/Input/index.module.css
RicardoVaz1: Ok, I'll remove it

frontend/src/components/Input/index.js
RicardoVaz1: Yes, I changed to:
```
<div className={!error ? "" : styles.errorContainer}>
      {!error ? "" : <span>{errorMessage}</span>}
      <input
        disabled={disabled}
        type={type}
        placeholder={placeholder}
        className={
          !error ? styles.input : `${styles.input} ${styles.inputError}`
        }
      />
</div>
```

frontend/src/components/Input/index.module.css
RicardoVaz1: Ok, I'll remove it for now, and I'll ask Hannah the designs for focused/active state

frontend/src/components/Input/index.module.css
RicardoVaz1: Done

frontend/src/components/Input/index.module.css
RicardoVaz1: I fixed the first topic. Should I change this:
```
<div className={!error ? "" : styles.errorContainer}>
      {!error ? "" : <span>{errorMessage}</span>}
      <input
        disabled={disabled}
        type={type}
        placeholder={placeholder}
        className={
          !error ? styles.input : `${styles.input} ${styles.inputError}`
        }
      />
</div>
```
to an `Error` component?

frontend/src/components/Input/index.js
RicardoVaz1: Done

frontend/src/components/Input/index.js
RicardoVaz1: Done

frontend/src/components/Input/index.js
RicardoVaz1: I removed the texts object and I added this:
```
Input.defaultProps = {
  disabled: false,
};
```

joaojusto (approved):
LGTM, as soon as @mirhamasala is happy 🚢

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,44 @@
+.errorMessage {
+  padding: 0;
+  margin: 0;
+  margin-bottom: var(--font-size-xxs);
```
mirhamasala, line 4: The variable is called `font-size` but we're using it for `margin` too?

mirhamasala, line 4: We're using the `margin` shorthand to reset all values to `0` and then we add `margin-bottom` shorty after.

Also, the default styles for a `p` element only have `margin-bottom` and `margin-top` applied so I wouldn't use a shorthand to reset those.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,44 @@
+.errorMessage {
+  padding: 0;
```
mirhamasala, line 2: A default `p` element doesn't have any `padding` applied so there's no need for this declaration.

frontend/src/components/Input/index.module.css
mirhamasala: I would also nudge us to think of a reset/normalize since we'll be dealing with a lot more `p` elements in the project.

mirhamasala, line 4:
```suggestion
  margin-top: 0;
  margin-bottom: var(--spacing-xxs); // this var doesn't exist yet
```

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,44 @@
...
+  margin: 0;
+  margin-bottom: var(--font-size-xxs);
+  font-size: var(--font-size-xxs);
+  line-height: var(--font-size-xs);
+  color: var(--color-red);
+}
+
+.input {
+  padding: var(--font-size-sm) var(--font-size-md);
```
mirhamasala, line 11: Same: we're using `--font-size` vars for `padding`

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,44 @@
+.errorMessage {
+  padding: 0;
+  margin: 0;
+  margin-bottom: var(--font-size-xxs);
+  font-size: var(--font-size-xxs);
+  line-height: var(--font-size-xs);
+  color: var(--color-red);
+}
+
+.input {
```
mirhamasala, line 10: Also, this is why I suggested to first look into adding `reset` or `normalize` since a lot of these styles would be handled already.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,44 @@
...
+  padding: var(--font-size-sm) var(--font-size-md);
+  font-family: var(--font-family-body);
+  font-size: var(--font-size-xs);
+  font-style: normal;
+  font-weight: 400;
```
mirhamasala, line 15: These are the defaults

```suggestion
```

frontend/src/styles/globals.css
```diff
@@ -13,8 +13,10 @@
   --color-gray: #808080;
   --color-light-gray: #bfbfbf;
   --color-white: #fff;
+  --color-red: #f73f52;
   --font-family-body: "Work Sans", sans-serif;
   --font-family-title: "Spicy Rice", cursive;
+  --font-size-xxs: 12px;
```
mirhamasala, line 19: We aren't using this for `font-size` but padding and margins

mirhamasala (commented):
Looking much better! ✨ Have a look at my suggestions. Make the necessary changes if you agree, and let's get this merged!

frontend/src/styles/globals.css
joaojusto: @mirhamasala this is a bit out of scope for this PR, we will rename this constants to better fit usage if you agree

frontend/src/components/Input/index.module.css
joaojusto: Beatriz just added them in a separated PR, we can rebase as soon as she merges it @RicardoVaz1

frontend/src/components/Input/index.module.css
joaojusto: Check my other reply, this was me, we'll do a follow up renaming the variables

joaojusto: I disagree with you here @mirhamasala, why do you think it's better to reset them more verbosely than just using the shorthand?

mirhamasala: Something I learned and find useful, explained here: https://codeguide.co/#shorthand-notation

mirhamasala: And I think this commit kinda proves that point, right?

We had both:

```
margin: 0;
margin-bottom:  var(--font-size-xxs);
```

frontend/src/styles/globals.css
mirhamasala: I agree, and in that case, we should just use `12px` directly and remove this var.

mirhamasala:
Looking at design, I don't think the placeholder text should be red when there's an error message.

Screenshot 2022-08-03 at 13 44 02

frontend/src/components/Input/index.module.css
joaojusto: The argument is that it leads to slopier code and unintended side effects, I don't see how that applies here, especially since you understood right away that it was supposed to be a margin reset. What side effects should we be expecting?

frontend/src/styles/globals.css
joaojusto: Or we can leave it as I suggest Ricardo and then simply rename these variables across the project. Seems less work than renaming them and then going to each individual component to start using them, or am I missing something?

frontend/src/components/Input/index.module.css
mirhamasala: I didn't understand straight away, actually–and that's the thing.

mirhamasala: Besides, if you ever need to add `right` and `left` spacing, this makes even less sense.

frontend/src/components/Input/index.module.css
```diff
@@ -0,0 +1,44 @@
...
+.input:disabled {
+  color: var(--color-dark-gray);
+  background-color: var(--color-light-gray);
+  border: 1px solid var(--color-dark-gray);
+  box-shadow: 0 4px 5px var(--color-gray);
+}
+
+::placeholder {
+  color: var(--color-light-gray);
+}
```
mirhamasala, line 32: Thought I left a comment regarding this. I believe this will leak into other components if we don't use `.input`

frontend/src/styles/globals.css
mirhamasala: I don't see how using `12px` directly isn't better? This is incorrect.

frontend/src/components/Input/index.module.css
joaojusto: I fail to see how `margin-top: 0` is supposed to be more explicit about the intention of the rule than `margin: 0` but maybe we should just disagree here.

frontend/src/components/Input/index.module.css
mirhamasala: So I believe this style will leak into other components if we don't prefix it with `.input`

frontend/src/components/Input/index.module.css
mirhamasala: `margin: 0` is clear (although, again, unnecessary, because you only want to reset the `top`)

But `margin: 0` followed with `margin-bottom: 12px` is more confusing than `margin-top: 0` `margin-bottom: 12px`

frontend/src/styles/globals.css
mirhamasala: I misunderstood. I meant adding `spacing` variables, not renaming the existing `font-size` ones.

PR #49, opened 2022-08-01
Add lead table and migrations

beasrodrigues24:
Why:
* For our app, we will need to store user data
* We want to associate each lead to a prize they have won, after they have played

How:
* By creating a Lead model that validates the presence of the mandatory values
* By creating a migration that creates a lead
* By creating a migration that adds a column that represents a key that matches the prize the person has won (it's null until they have played) - consulted https://guides.rubyonrails.org/active_record_migrations.html#foreign-keys

### DIFF
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -1,0 +1,3 @@
+class Lead < ApplicationRecord
+    validates :name, :email, :interests, presence: true
+end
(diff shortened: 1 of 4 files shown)


beasrodrigues24:
SIde note: decided that data_proc_consent and delivered aren't mandatory parameters. If it's created without them it's assumed they are false. Rails hates that we pass something with false on it (just doesn't seed it at all) so I did it this way. Thought that it doesn't even make sense to set delivered as true from the beginning anyways because delivered will be updated only after awarding the prize and picking it up.

joaojusto (approved):
🚢

PR #50, opened 2022-08-01
Add CRUD for leads

beasrodrigues24:
Why:
* The front-end app should be able to make post request to the back-end to add new leads

How:
* By adding a new route
* By adding the create endpoint
* By testing through postman if things are functional

### DIFF
--- a/app/controllers/leads_controller.rb
+++ b/app/controllers/leads_controller.rb
@@ -1,0 +1,11 @@
+class LeadsController < ApplicationController
+
+    def create 
+        lead = Lead.create(leads_params)
+        render json: lead
+    end 
+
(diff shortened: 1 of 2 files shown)


beasrodrigues24:
A couple of notes: this PR will be rebased once my other PR creating the tables is merged to main. This PR cannot be merged before that one. So it is also not a priority in being reviewed although I'd like some notes if anyone can. I didn't do any show endpoint as requested in the task because I was thinking that should only exist behind maybe active admin but might be confused.

Other thing: I've tested the post request through postman and it seems to be working well I'll leave a picture here.
image

joaojusto (approved):
LGTM

PR #51, opened 2022-08-02
Add rubocop to CI workflow

beasrodrigues24:
Why:
* Rubocop wasn't being used the CI workflow

How:
* Adding a new Github Actions workflow to run Rubocop
* Fixing all the warnings
* Adding a `.rubocop.yml` configuration file

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,0 +1,21 @@
+name: Backend Lint
+
+on:
+  pull_request:
+    branches: ["main"]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy: 
+      matrix: {dir: ["app", "config"]}
+
+    steps: 
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.1.2'
+          bundler-cache: true
+      - run: bundle exec rubocop
+        working-directory: ${{ matrix.dir }}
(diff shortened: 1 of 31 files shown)


joaojusto (approved):

DavideSilva:
We can ignore everything that is in the `db` and `bin` folders from Rubocop. We don't want to lint the `bin` folder since it's mostly scripts and might not be Ruby code and we don't need to lint the `db` folder since files in that directory are mostly generated and we don't want to mess with that

beasrodrigues24:
> We can ignore everything that is in the `db` and `bin` folders from Rubocop. We don't want to lint the `bin` folder since it's mostly scripts and might not be Ruby code and we don't need to lint the `db` folder since files in that directory are mostly generated and we don't want to mess with that

Currently working on that. I applied rubocop to app, config and db but with what you said now I'll amend the last commit to only apply it to app and config. Sounds good?

DavideSilva:
Yep, but we need to have a Rubocop configuration file to tell Rubocop to ignore those files. Otherwise, when we run this in the CI, Rubocop will try to lint those folders either way.

I'll push a commit to add that config file

beasrodrigues24:
Okay, thanks :smile:

beasrodrigues24:
Is it good to go or should I fix something else?

DavideSilva:
We should ignore the changes that Rubocop made to these folders that are already committed to this PR. We don't want to change things in the `db` folder

beasrodrigues24:
And what is the best way to do that? Right now the only way I see myself being able to do that is to just go to the files and putting them back as I remembered them to be (without the comments on top) but I assume there might be a more proper way.

DavideSilva:
There is. I can show you if you want to. Let's jump on a call, it will be easier that way

beasrodrigues24:
Yes, please. Huddle?

PR #52, opened 2022-08-02
Add heading 3 component

jreis13:
Why:

How:
* Add and style the heading 3 component, using variables from constant file.


joaojusto (approved):
@jreis13 Same thing for Heading1 and Heading2

PR #53, opened 2022-08-02
Add heading 2 component

jreis13:
Why:

How:
* Add and style the heading 2 component, using variables from constant file.


joaojusto (approved):
@jreis13 I have the same comments I left on Heading1.

PR #54, opened 2022-08-02
Add heading 1 component

jreis13:
Why:

How:
* Add and style the heading 1 component, using variables from constant file.


frontend/src/components/Typography/Head1/index.module.css
```diff
@@ -0,0 +1,6 @@
+.head1 {
+  font-family: var(--font-family-body);
+  font-size: var(--font-size-lg);
+  font-weight: 500;
+  line-height: 40px;
+}
```
joaojusto, line 6: Aren't we missing `color` here? 🤔

frontend/src/components/Typography/Head1/index.module.css
```diff
@@ -0,0 +1,6 @@
+.head1 {
```
joaojusto, line 1: Would you mind keeping `heading1` to be consitent with design naming?

joaojusto (approved):
@jreis13 This is good, just take a look at my comments but feel free to merge as soon as you tackle them

frontend/src/components/Typography/Head1/index.module.css
jreis13: Added

frontend/src/components/Typography/Head1/index.module.css
jreis13: Makes sense, changed in this as well as in the other heading components.

PR #55, opened 2022-08-02
Add logo component

jreis13:
Why:

How:
* Add and style the title component, using variables from constant file.


joaojusto (commented):
@jreis13 I'm also not sure about the element we should use for this one. Which element are we using for Heading1? h1 as well?

jreis13:
> @jreis13 I'm also not sure about the element we should use for this one. Which element are we using for Heading1? h1 as well?

Yes we are using h1 as well,  I was unsure what would be the best element to use.

joaojusto (approved):

PR #56, opened 2022-08-02
Add subtitle component

jreis13:
Why:

How:
* Add and style the subtitle component, using variables from constant file.

### DIFF
--- a/frontend/src/components/Typography/Subtitle/index.js
+++ b/frontend/src/components/Typography/Subtitle/index.js
@@ -1,0 +1,12 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.css";
+
+const Subtitle = ({ children }) => {
+  return <h2 className={styles.subtitle}>{children}</h2>;
+};
+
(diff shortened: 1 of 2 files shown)


frontend/src/components/Typography/Subtitle/index.js
```diff
@@ -0,0 +1,12 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.css";
+
+const Subtitle = ({ children }) => {
+  return <h2 className={styles.subtitle}>{children}</h2>;
```
joaojusto, line 5: I'm not sure this should be an H2, what's the difference between this one and Heading2 component?

joaojusto (changes requested):

frontend/src/components/Typography/Subtitle/index.js
jreis13: There are some style differences major one being the font family. If not an h2, which element would it be more suitable to use?

PR #57, opened 2022-08-02
Add button component

jreis13:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5183280636
* Need to create the button component to use throughout the app.

How:
* Add and style the button component, using variables from constant file.


frontend/src/components/Button/index.module.css
```diff
@@ -1,12 +1,16 @@
...
-  padding: 15px 40px;
-  margin-top: 64px;
-  font-family: "Libre Baskerville", serif;
-  font-size: 20px;
-  background: #d9d9d9;
-  border: 1px solid #000;
-  border-radius: 5px;
-}
+  width: 335px;
+  height: 60px;
```
joaojusto, line 3: We don't want fixed with/height buttons. The button width should be fluid, meaning it's size is a result of the space needed to render whatever text content we have in there plus the padding around it

joaojusto (changes requested):

joaojusto:
@jreis13 Please update the PR title and link the corresponding task please. See my message on slack regarding it.

joaojusto (approved):
LGTM 🚢

PR #58, opened 2022-08-02
Add missing association in Lead model

beasrodrigues24:
Why:
* Each lead will eventually be associated with the prize they have won (although that starts null), so the model should be a reflection of that

How:
* By adding has_one to the lead model, representing that each lead will be associated with one prize


beasrodrigues24:
Was unsure if I had to put a belongs_to because prizes can exist independent from being attributed to leads or not. Should I have used it?

DavideSilva:
What's the plan with the Prizes? Do we want to have one Prize belong to one Lead and have multiple of the same Prize for the wheel or do we want to one have Prize and be able to assign that Prize to multiple Leads?

Basically what I'm asking is: what's the relation between Prizes and Leads? Is it 1:1 or 1:n?

beasrodrigues24:
So we want to be multiple prizes and those prizes will be fetched by the wheel. It's necessary to associate one prize to each lead when they have won. At the start it's null but then we associate the won prize to the user. So my answer is the same prize can be assigned to multiple leads, but a lead can only have one prize associated (and that prize can be null for a while).

beasrodrigues24:
So you said that I should put a belongs_to prize in the Lead model. However I encountered an issue: in the presence of a belong_to rails forces me to when creating a Lead to attribute it a prize immediately. It doesn't let me create a lead without a prize associated (which is necessary for the way we are planning the app for now). Quoting from what I read "belong_to associations have an implicit presence validation on them (that's what required: true does under the hood". So I commited without it and tomorrow we can talk about this again ! It works the way it is right now though

db/migrate/20220802162516_update_prize_lead_foreign_key.rb
```diff
@@ -0,0 +1,5 @@
+class UpdatePrizeLeadForeignKey < ActiveRecord::Migration[7.0]
+  def change
+    rename_column :leads, :prizes_id, :prize
```
DavideSilva, line 3: The renamed column should be `:prize_id`. That way Rails knows how to do its magic with the associations

This is probably why Rails was complaining when you tried to add the `belongs_to` in the `Lead` model

DavideSilva (changes requested):

db/migrate/20220802162516_update_prize_lead_foreign_key.rb
beasrodrigues24: image
It still gives the same error even after doing this fix.

DavideSilva: You can add an `optional: true` to the `belongs_to` statement

image

beasrodrigues24: That makes sense, thanks!!

DavideSilva (approved):
:shipit:

PR #59, opened 2022-08-02
Set up mailcatcher config

beasrodrigues24:
Why:
* The app will need a service to send email

How:
* Completing the development file in environments directory with information about the action mailer. Followed the recomendations from https://mailcatcher.me/ . This included not putting mailcatcher in Gemfile


beasrodrigues24:
This feels a bit empty but for configuration it seems to only be this it has to be done for the config part. Will do this part in another PR 
image

DavideSilva (approved):

PR #62, opened 2022-08-03
Add CSS normalize

beasrodrigues24:
Why:
* It's very helpful while writing CSS. To see what it does, consult https://github.com/necolas/normalize.css
* Task: https://3.basecamp.com/4319812/buckets/26581088/todos/5192137049

How:
* By running npm install --save normalize.css
* By adding @import-normalize to index.css (Note: if VSC complains about this import do change the css.lint.unknownAtRules setting to ignore)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,6 +13,7 @@
         "@testing-library/user-event": "^13.5.0",
         "axios": "^0.27.2",
         "classnames": "^2.3.1",
+        "normalize.css": "^8.0.1",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-router-dom": "^6.3.0",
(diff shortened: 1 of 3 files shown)


joaojusto (changes requested):
I think you forgot to import `normalize.css` somewhere, otherwise it won't be included

joaojusto:
@beasrodrigues24 please link the task on the PR description, and assign yourself to the task so that no one else starts doing it and we duplicate work. Thanks

beasrodrigues24:
On it!

beasrodrigues24:
So I added to the index.css more external to the app so it  inherits to the rest of the app. VSC will complain about it but after reading through it I left the way they recommend to bypass that complaint in the PR description

joaojusto (approved):
As soon as you make sure this is working :shipit:

PR #64, opened 2022-08-03
Update prize table

jreis13:
Why:
* Need to add fields to the prize table.

How:
* Replace `is_merch` for `type`
* Add `description`
* Add `identifier`

### DIFF
--- a/app/controllers/prizes_controller.rb
+++ b/app/controllers/prizes_controller.rb
@@ -23,6 +23,6 @@ class PrizesController < ApplicationController
   private
 
   def prize_params
-    params.require(:prize).permit(:name, :percentage, :image, :is_merch)
+    params.require(:prize).permit(:name, :percentage, :image, :prize_type, :description, :identifier)
   end
 end
--- a/app/models/prize.rb
+++ b/app/models/prize.rb
@@ -4,5 +4,8 @@ class Prize < ApplicationRecord
   validates :name, presence: true
   validates :percentage, presence: true
   validates :image, presence: true
+  validates :prize_type, presence: true
+  validates :description, presence: true
+  validates :identifier, presence: true
   has_many :leads
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -9,6 +9,6 @@
 # bin/rails c    OR   rake db:reset AND bin/rails db:seed
 # rake db:drop db:create db:migrate db:seed
 5.times do |prize|
-  Prize.create(name: "Tshirt #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/t-shirt.svg", is_merch: true)
-  Prize.create!(name: "NFT #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/nft.svg")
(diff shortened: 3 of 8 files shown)


joaojusto (changes requested):
Changing migrations is not recommended. This will break the database on the other computers and had we some production database deployed it would break it as well.
I would recommend you to generate a new migration to change the prize table accordingly

jreis13:
> Changing migrations is not recommended. This will break the database on the other computers and had we some production database deployed it would break it as well.
> I would recommend you to generate a new migration to change the prize table accordingly

Understood, fixed it now.

app/controllers/prizes_controller.rb
```diff
@@ -23,6 +23,6 @@ def win_prize
   private
 
   def prize_params
-    params.require(:prize).permit(:name, :percentage, :image, :is_merch)
+    params.require(:prize).permit(:name, :percentage, :image, :prize_type, :description, :identifier)
```
mirhamasala, line 26: Shouldn't this be `type` now?

app/models/prize.rb
```diff
@@ -4,5 +4,8 @@ class Prize < ApplicationRecord
   validates :name, presence: true
   validates :percentage, presence: true
   validates :image, presence: true
+  validates :prize_type, presence: true
```
mirhamasala, line 7: Same as above

db/seeds.rb
```diff
@@ -9,6 +9,6 @@
 # bin/rails c    OR   rake db:reset AND bin/rails db:seed
 # rake db:drop db:create db:migrate db:seed
 5.times do |prize|
-  Prize.create(name: "Tshirt #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/t-shirt.svg", is_merch: true)
-  Prize.create!(name: "NFT #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/nft.svg")
+  Prize.create(name: "Tshirt #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/t-shirt.svg", prize_type: 'Merch', description: 'This is a merch item', identifier: 'merch')
+  Prize.create!(name: "NFT #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/nft.svg", prize_type: 'NFT', description: 'This is a NFT item', identifier: 'NFT')
```
mirhamasala, line 13: Same as above

app/controllers/prizes_controller.rb
jreis13: Was getting an error and figured that type is a reserved keyword I believe, used `prize_type` instead in this case.

mirhamasala: And it's okay to have `type` in the migration?

mirhamasala: Haven't done Rails for a while. 😅 Did you have to roll back? And create a new migration?

mirhamasala:
Let's have @DavideSilva or @zediogoviana have the final check on this one before merging!

DavideSilva (approved):
Looks okay :shipit:

mirhamasala (approved):

PR #65, opened 2022-08-03
Change font-size to size variables

jreis13:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5192574897
* We were using `font-size` prefix for the sizings but we should change them to be more generic. 

How
* Use the prefix `size` (ie: `size-xs` instead of `font-size-xs`).

### DIFF
--- a/frontend/src/components/Button/index.module.css
+++ b/frontend/src/components/Button/index.module.css
@@ -2,7 +2,7 @@
   min-width: 335px;
   padding: 20px;
   font-family: var(--font-family-body);
-  font-size: var(--font-size-xs);
+  font-size: var(--size-xs);
   font-weight: 600;
   line-height: 20px;
   color: var(--color-blue);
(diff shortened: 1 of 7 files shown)


mirhamasala:
@joaojusto I understand the confusion now. What I meant was to add `--spacing-xs` etc. not to rename the `font-size` ones.

joaojusto:
@mirhamasala in this case it's the same, Hanna is using typography as base and all the spacings are based on the typographic scale, so I think it's safe for us to share

mirhamasala:
> @mirhamasala in this case it's the same, Hanna is using typography as base and all the spacings are based on the typographic scale, so I think it's safe for us to share

What happens when we need to add  2, 4, or `6px` or `128px`? Like values that can't be used for `font-size`?

joaojusto:
@mirhamasala nothing, either we use those values directly, as in the case for box-shadows, borders and other elements/rules which are completely disconnected from the typographic scale or we use fractions (`size-xs / 2`).

Although most of the internet says otherwise I'd argue very hard against having all values codded into variables.

mirhamasala:
@joaojusto Then, shall we keep the the `font-size` variables and for everything else etc. just use the values directly? It would make this PR redundant.

PR #67, opened 2022-08-03
Send email with the prize to the Lead

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5183373573

How:
* By following https://guides.rubyonrails.org/action_mailer_basics.html
* Using Mailcatcher

### DIFF
--- a/app/controllers/prizes_controller.rb
+++ b/app/controllers/prizes_controller.rb
@@ -23,7 +23,9 @@ class PrizesController < ApplicationController
     prize_info = Prize.find_by(name: win_prize)
     lead.update(prize: prize_info)
 
-    render json: Prize.find_by(name: win_prize)
+    ApplicationMailer.with(email: params[:email], name: lead).win_prize_email.deliver_now
+
+    render json: Prize.find_by(name: pickup.pick)
   end
 
   private
--- a/app/controllers/prizes_controller.rb
+++ b/app/controllers/prizes_controller.rb
@@ -16,16 +16,18 @@ class PrizesController < ApplicationController
       acc[prize.name] = prize.percentage
     end
 
-    pickup = Pickup.new(prizes_list)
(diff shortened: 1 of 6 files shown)


DavideSilva (changes requested):
What's the point in having rubocop if we are just going to ignore the warnings and disable the rules? 😂

app/controllers/prizes_controller.rb
```diff
@@ -11,20 +11,30 @@ def create
     render json: prize
   end
 
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
   def win_prize
```
DavideSilva, line 16: Looking at this method, it seems that what we are doing is basically:
- sort a new prize
- assign the prize to the lead
- send an email to the lead

So, to fix the warnings about the method being too long and doing way too much complex things, I'd like you to refactor the `win_prize` method into a method that just calls three other methods. Something like:
```
def win_prize
  sort_prize
  assign_prize_to_lead
  send_email

  render ...
end
```

RicardoVaz1: I fixed the win_prize method, it's now calling the other methods, which are returning values that are being passed to the next methods. Is it a valid solution?

DavideSilva: Not only is it a valid solution, it's also more idiomatic Rails. 

It's very common to have classes in Rails that have the public methods be mostly calls to private methods. It helps keep code clear and concise and forces us to name chunks of code that otherwise might be tricky to understand their purpose.

It's easier for the reader to understand what the `win_prize` method does by reading the private methods names than trying to understand what the entire implementation does

Good job 👍

DavideSilva (approved):
:shipit:

PR #68, opened 2022-08-04
Add checkbox component

beasrodrigues24:
Why:
* Task: https://3.basecamp.com/4319812/buckets/26581088/todos/5183284984

How:
* By using combination of input, label, span and css

### DIFF
--- a/frontend/src/components/Checkbox/index.js
+++ b/frontend/src/components/Checkbox/index.js
@@ -1,0 +1,51 @@
+import { PropTypes } from "prop-types";
+
+import styles from "./index.module.css";
+
+function Checkbox({ id, label, disabled, error, errorMessage }) {
+  return (
+    <>
+      {error ? <p className={styles.errorMessage}>{errorMessage}</p> : null}
+      <div className="inputWrapper">
+        <input
+          id={id}
+          className={styles.input}
+          type="checkbox"
+          disabled={disabled}
+        />
+        <label htmlFor={id} className={styles.label}>
+          {label}
+        </label>
+        <span className={styles.control}>
+          <svg
+            viewBox="0 0 10 8"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path
+              d="M1 4L4 6.66667L9 1.33333"
+              stroke="currentColor"
+              strokeWidth="2"
+              strokeLinecap="round"
+              strokeLinejoin="round"
+            />
+          </svg>
+        </span>
+      </div>
+    </>
+  );
+}
+
+Checkbox.propTypes = {
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  disabled: PropTypes.bool,
+  error: PropTypes.bool,
+  errorMessage: PropTypes.string,
+};
+
+Checkbox.defaultProps = {
+  disabled: false,
+};
+
+export default Checkbox;
--- a/frontend/src/components/Checkbox/index.module.css
+++ b/frontend/src/components/Checkbox/index.module.css
@@ -1,0 +1,45 @@
+.inputWrapper {
+  --size-checkbox: 24px;
+
+  display: flex;
+  align-items: center;
+}
+
+.input {
+  width: var(--size-checkbox);
+  height: var(--size-checkbox);
+  opacity: 0;
+}
+
+.label {
+  padding-left: 20px;
+  font-size: var(--font-size-sm);
+  color: var(--color-blue);
+}
+
+.control {
+  position: absolute;
+  left: 0;
+  display: inline-grid;
+  place-items: center;
+  width: var(--size-checkbox);
+  height: var(--size-checkbox);
+  color: transparent;
+  pointer-events: none;
+  cursor: pointer;
+  border: 2px solid var(--color-blue);
+  border-radius: 4px;
+}
+
+.control > svg {
+  width: 12px;
+}
+
+.input:checked ~ .control {
+  color: var(--color-blue);
+}
+
+.errorMessage {
+  font-size: var(--font-size-xxs);
+  color: var(--color-red);
+}


frontend/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,29 @@
+import { PropTypes } from "prop-types";
+
+import styles from "./index.module.css";
+
+function Checkbox({ id, children }) {
```
mirhamasala, line 5: For the label text, I'd just use the prop `label` instead of `children`

The usecase is a bit different than for `button` and `h` headings

frontend/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <>
+      <input id={id} className={styles.input} type="checkbox"></input>
```
mirhamasala, line 8: No need for a closing tag

frontend/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,29 @@
...
+      <label htmlFor={id}>{children}</label>
+      <span className={styles.control}>
+        <svg viewBox="0 0 24 24" aria-hidden="true" focusable="false">
+          <path
+            fill="none"
+            stroke="currentColor"
+            strokeWidth="4"
+            d="M1.73 12.91l6.37 6.37L22.79 4.59"
+          />
+        </svg>
```
mirhamasala, line 18: Is this the SVG that I used in de CodeSandbox?

frontend/src/components/Checkbox/index.js
```diff
@@ -0,0 +1,29 @@
...
+      </span>
+    </>
+  );
+}
+
+Checkbox.propTypes = {
+  id: PropTypes.string.isRequired,
+  children: PropTypes.string.isRequired,
```
mirhamasala, line 26: You're missing a `disabled` prop. Keep in mind that you can set default states as well. 

https://reactjs.org/docs/typechecking-with-proptypes.html

frontend/src/components/Checkbox/index.module.css
```diff
@@ -0,0 +1,33 @@
+.label {
+  color: var(--color-black);
+  pointer-events: none;
+}
+
+.input {
+  width: var(--font-size-md);
+  height: var(--font-size-md);
```
mirhamasala, line 8: Since this `width` and `height` are coupled with the `width` and `height` of the `.control` element, it would be nice if you could create local variables and use those. Because a developer that comes after you might not realize that those two are coupled and might only change the size in one place, effectively breaking the feature,

frontend/src/components/Checkbox/index.module.css
```diff
@@ -0,0 +1,33 @@
+.label {
+  color: var(--color-black);
+  pointer-events: none;
```
mirhamasala, line 3: What does this do?

mirhamasala:
This is what I'm seeing. Is this correct? Because the text doesn't seem center-aligned like in the design.

Screenshot 2022-08-04 at 14 11 13

mirhamasala:
I'm also missing the error text as per design. See `input` component.

frontend/src/components/Checkbox/index.module.css
mirhamasala: You can declare something like this at the top of the file:

```
--size-checkbox: 24px;
```

Then:

```
width: var(--size-checkbox);
height: var(--size-checkbox);
```

beasrodrigues24:
The error text was confusing to me because I don't really now what to do because it will be mandatory to select a checkbox in the interests page but when the checkbox is used for the data processing consent it is not mandatory.

frontend/src/components/Checkbox/index.module.css
mirhamasala: You can also use the global variables if you prefer. I still don't agree with using the `font-size` one here though, but it would look something like this:

```
---size-checkbox: var(---font-size-md)
```

Then

```
width: var(--size-checkbox)
height: var(--size-checkbox)
```

frontend/src/components/Checkbox/index.js
beasrodrigues24: Will change it then. Gonna export it from Hannah's design

mirhamasala:
> The error text was confusing to me because I don't really now what to do because it will be mandatory to select a checkbox in the interests page but when the checkbox is used for the data processing consent it is not mandatory.

Maybe you need some logic that checks if the checkbox is `required`

frontend/src/components/Checkbox/index.js
beasrodrigues24: I don't understand what you mean with disable prop :/

frontend/src/components/Checkbox/index.module.css
beasrodrigues24: That makes sense!

frontend/src/components/Checkbox/index.module.css
beasrodrigues24: That should've been color purple but now that I see it doesn't even do anything so I'll rethink all of this. The part of pointer-events was because I kept pressing the label and it checked and unchecked the box and I thought that was odd.

frontend/src/components/Checkbox/index.module.css
beasrodrigues24: So it gives me an error unless I put it inside something. I that I could put it inside :root , is that correct?

frontend/src/components/Checkbox/index.module.css
beasrodrigues24: Turns out I'm an airhead and I forgot putting the className on <label> :)

frontend/src/components/Checkbox/index.module.css
mirhamasala: Putting it inside `:root` will give you access to this variable in your entire project. You only need it locally. Try adding it under the top class.

frontend/src/components/Checkbox/index.module.css
mirhamasala: The label pressing thing is a [feature](https://css-tricks.com/html-inputs-and-labels-a-love-story/#:~:text=There%20are%20two%20ways%20to,input%20and%20holding%20its%20hand.) not a bug.. keep it! (For your info, it's achieved by the `for` and `id`.

frontend/src/components/Checkbox/index.js
mirhamasala: All input elements can also be `disabled` > [Check](https://www.geeksforgeeks.org/html-dom-input-checkbox-disabled-property/#:~:text=A%20disabled%20checkbox%20is%20unclickable,the%20Input%20Checkbox%20disabled%20property.)

mirhamasala: You can also check out Ricardo's `input` element.

beasrodrigues24: Will do!

beasrodrigues24:
So there is still some stuff I think I could do better but I think that it's ready for another review. Think I did everything that was mentioned!

mirhamasala (approved):
I added some little changes. Most importantly, the text still wasn't center-aligned, so I added the appropriate styles under `.inputWrapper`.

I also increased the size of the checkmark a bit because even though you used the correct size per design, it turned out quite small.

I also removed the red border on error because that's not present in the design.

On a side note, and for a future refactor, it seems to me that `errorMessage` should be a component of its own. Because repeating.

Screenshot 2022-08-04 at 19 56 18

Please fixup the last two commits, and let's get this merged. Great job! 🚀

beasrodrigues24:
I'll fix up tomorrow morning because I might need some help doing it because I've never done that before. I assume I have to rebase on my own branch and then just put f next to the CR commits but I'm unsure if I can create any issues after that :sweat_smile:

PR #73, opened 2022-08-05
ActiveAdmin setup

DavideSilva:
Why:
* We need to provide some sort of admin functionality
* Following [this](https://jontzavala.medium.com/setting-up-active-admin-with-rails-6-api-6232c22c3d36) to make the changes needed to support ActiveAdmin in a Rails API app 

How:
- Adding ApiController and update existing controllers to use it
- Updating ApplicationController to inherit from ActionController::Base instead
- Adding ActiveAdmin and required gems
- Updating the `application/config.rb` to include the necessary middlewares
- Running the ActiveAdmin generator, `rails g active_admin:install`
- Running migrations


joaojusto (approved):
LGTM

PR #74, opened 2022-08-05
Add more props to input component

beasrodrigues24:
Why:
* In the purposes of doing the sign up page, noticed that it wouldn't be possible to get the information on the box without adding a few props

How:
* By adding onChange, value and an id to the input custom component

### DIFF
--- a/frontend/src/components/Input/index.js
+++ b/frontend/src/components/Input/index.js
@@ -2,7 +2,16 @@ import { PropTypes } from "prop-types";
 import styles from "./index.module.css";
 import classNames from "classnames";
 
-const Input = ({ type, placeholder, disabled, error, errorMessage }) => {
+const Input = ({
+  id,
+  type,
+  placeholder,
+  disabled,
+  error,
+  errorMessage,
+  value,
+  onChange,
+}) => {
   const inputClassNames = classNames(styles.input, {
     [styles.hasError]: error,
   });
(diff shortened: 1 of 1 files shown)


frontend/src/components/Input/index.js
```diff
@@ -2,7 +2,16 @@ import { PropTypes } from "prop-types";
...
 
-const Input = ({ type, placeholder, disabled, error, errorMessage }) => {
+const Input = ({
+  id,
+  type,
+  placeholder,
+  disabled,
+  error,
+  errorMessage,
+  value,
```
mirhamasala, line 12: Why is `value` a prop? Is the parent passing down this?

beasrodrigues24: The parent will. My idea is that the signup page will have `const [name, setName] = useState("")` and I'll pass value={name} and onChange={(event) => setName(event.target.value)}.

mirhamasala: Are you trying to pass the `value` back to the parent? [Lifting state up](https://reactjs.org/docs/lifting-state-up.html)?

beasrodrigues24: Then the same for email I assume

beasrodrigues24: I decided to push this first before the Signup page to build on a branch created out of this branch

beasrodrigues24: Yes I think I am (only saw that message now after updating the page )

mirhamasala: Did you try this?

beasrodrigues24: Yes, and it worked. Maximino helped me. See [thread](https://subvisualuniverse.slack.com/archives/C03KFP6GQR3/p1659695443983429)

mirhamasala: Ah, I see!

mirhamasala: Got it.

mirhamasala (approved):

PR #75, opened 2022-08-05
Add Sign Up page

beasrodrigues24:
Why:
    * Task: https://3.basecamp.com/4319812/buckets/26581088/todos/5183295453
    
How:
    * Assembling components created to be used in this page

### DIFF
--- a/frontend/src/components/Checkbox/index.js
+++ b/frontend/src/components/Checkbox/index.js
@@ -2,7 +2,7 @@ import { PropTypes } from "prop-types";
 
 import styles from "./index.module.css";
 
-function Checkbox({ id, label, disabled, error, errorMessage }) {
+function Checkbox({ id, label, disabled, onChange, error, errorMessage }) {
   return (
     <>
       {error ? <p className={styles.errorMessage}>{errorMessage}</p> : null}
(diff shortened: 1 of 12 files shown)


beasrodrigues24:
I think the first commit could be cherry picked to another PR. I'm gonna look on how to do that later. 
About the rest, there's some things that aren't exacly like the design that I'm gonna tweak later. Just wanted to push so it could be reviewed in the mean time!

beasrodrigues24:
This has a lot of CSS problems that I'm gonna have to ask for help to fix :/

joaojusto (changes requested):
This is almost there, I would: 
1 - Extract each "from" into a separated component"
2 - Use constants to switch between each state, ie: `const STEP_1 = "STEP_1"` or `const NAME_STEP = "NAME_STEP"`

beasrodrigues24:
I'm confused about the constants part. Do I create them and then use them in the useState? @joaojusto

beasrodrigues24:
The CSS is still not good at all, so if a pairing session is possible tomorrow it would be great :pray:

joaojusto (approved):

PR #78, opened 2022-08-08
Fix Wheel

RicardoVaz1:
Why: 
* https://3.basecamp.com/4319812/buckets/26581088/todos/5183296780

How:
* Fixing the Wheel page and component to look like the design
* Changing main and secondary colors
* Changing from 5.times to 6.times


RicardoVaz1:
We will only have 2 colors for the slices (instead of 3): #EA1889 and #9F18D9. And we'll have a verification on the database to only have even number of slices

joaojusto (changes requested):
I would move the prize page to the base router, and as soon as the animation ends navigate away to that page. We could store the won prize on `localStorage` or passe it as a url query argument.

joaojusto (approved):
:shipit:

PR #83, opened 2022-08-09
Fix connection between user info form and privacy policy

beasrodrigues24:
Why:
* The form information filled out by the user (that includes name, email and interests) wasn't being carried over to the privacy policy. So the POST request was filled with undefined values

How:
* By using local storage to carry over an object that includes that data


joaojusto (approved):
LGTM

PR #91, opened 2022-08-10
Add image to emails

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5216059191

How:
* Using a img tag with the src getting the image from the public folder


app/views/win_prize/win_merch.html.erb
```diff
@@ -8,17 +8,19 @@
   </head>
 
   <body>
+    <img src="<%= @server_url %>logoWheel.png" alt="Logo Wheel">
```
mirhamasala, line 11: Just dropping in for a detail: I'd namespace assets with a dash `logo-wheel.png`

RicardoVaz1: Ok, I'll fix it

RicardoVaz1: Done

app/mailers/application_mailer.rb
```diff
@@ -2,13 +2,14 @@
...
 class ApplicationMailer < ActionMailer::Base
-  default from: 'geral@subvisual.com'
+  default from: 'geral.subvisual@gmail.com'
 
   def win_merch_prize_email
     @lead_name = params[:lead_name]
     @prize_name = params[:prize_name]
     @code_to_claim = params[:code_to_claim]
     @url = 'mailto:contact@subvisual.com?subject=Hi%40Subvisual!%20I%20want%20to%20talk%20to%20you'
+    @server_url = 'https://subvisuwheel.netlify.app/'
```
zediogoviana, line 12: I think it was better if this was using an Env variable instead of having it hardcoded... Imagine you switch from using netlify to other services, or you change the name of the name under Netflify, or you simply want to use your custom domain, you'd need to change the code everywhere, also

RicardoVaz1: Ok, I'll fix it

RicardoVaz1: Done. I had to use `ENV.fetch['SERVER_URL']` because rubocop was complaining with `ENV['SERVER_URL']`

zediogoviana: Yep, makes sense. Have a look [here](https://stackoverflow.com/a/47985757/8733796) for a TLDR of the difference between both.

PR #95, opened 2022-08-10
Add disabled state to button

jreis13:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5215871363
* Design requires us to have a disabled button on the privacy policy page

How:
* Adding css rules so that we can display the button differently when its disabled

### DIFF
--- a/frontend/src/components/Button/index.js
+++ b/frontend/src/components/Button/index.js
@@ -1,5 +1,5 @@
 import { PropTypes } from "prop-types";
-import styles from "./index.module.css";
+import styles from "./index.module.scss";
 
 const Button = ({ children, href, onClick }) => {
   return href ? (
--- a/frontend/src/components/Button/index.module.scss
+++ b/frontend/src/components/Button/index.module.scss
@@ -14,4 +14,9 @@
   border: none;
   border-radius: 300px;
   box-shadow: 0 4px 3px var(--color-gray);
+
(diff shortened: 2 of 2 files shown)


frontend/src/components/Button/index.module.css
```diff
@@ -15,3 +15,8 @@
   border-radius: 300px;
   box-shadow: 0 4px 3px var(--color-gray);
 }
+
+.button:disabled {
+  color: var(--color-dark-gray);
+  background-color: var(--color-light-gray);
+}
```
mirhamasala, line 22: Why not use `sass` now that we have it installed?

```suggestion
  box-shadow: 0 4px 3px var(--color-gray);
  
  &:disabled {
    color: var(--color-dark-gray);
    background-color: var(--color-light-gray);
   }
}
```

mirhamasala: P.S. Not sure if you've tried this already, but if you agree, you can commit my suggestion straight from here. Look for the `Commit suggestion` button

mirhamasala (approved):
Minor suggestion to use SASS. ✨🚀

frontend/src/components/Button/index.module.css
jreis13: That's a good point, I'll do that 🙏

PR #96, opened 2022-08-10
Configure send emails

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5206155989

How:
* Using Action Mailer
* Following [this](https://medium.com/nerd-for-tech/implementing-action-mailer-ruby-on-rails-1766f59c6f) documentation


RicardoVaz1:
I created this branch from the brach `rv/add-images-to-emails` because I needed the changes in that branch, but doing that somehow it created the commit `Add image to emails` again, and I already have that commit in this [PR](https://github.com/Subvisual-Academy/subvisuwheel/pull/91), do I need to delete this commit?

mirhamasala:
> I created this branch from the brach `rv/add-images-to-emails` because I needed the changes in that branch, but doing that somehow it created the commit `Add image to emails` again, and I already have that commit in this [PR](https://github.com/Subvisual-Academy/subvisuwheel/pull/91), do I need to delete this commit?
> (Btw, I already added the env variables on the Heroku dashboard)

Once you merge #91, you can rebase this one. You can also set this one to "track" (not sure that's the right wording) #91 right from GitHub here. @DavideSilva showed @beasrodrigues24 once how to do it, but I don't remember how it was done.

config/environments/development.rb
```diff
@@ -56,9 +56,21 @@
...
-  config.action_mailer.raise_delivery_errors = false
+  # config.action_mailer.default_url_options = { :host => 'subvisuwheel.netlify.app', :protocol => 'https' }
+  config.action_mailer.smtp_settings = {
+    :address => 'smtp.gmail.com',
+    :port => 587,
+    :user_name => ENV["USER_NAME"],
+    :password => ENV["USER_PASSWORD"],
+    :authentication => 'plain',
+    :enable_starttls_auto => true
+  }
```
zediogoviana, line 73: Why are you removing the configuration for using Mailcatcher, and sending real emails? This is development, you can make some tests with the real thing, but it doesn't seem a good thing to have it every time. Mailcatcher should be enough for what you need

config/application.rb
```diff
@@ -43,5 +43,12 @@ class Application < Rails::Application
     config.middleware.use ActionDispatch::Flash
     config.middleware.use ActionDispatch::Cookies
     config.middleware.use ActionDispatch::Session::CookieStore
+
+    config.before_configuration do
+      env_file = File.join(Rails.root, 'config', 'local_env.yml')
+      YAML.load(File.open(env_file)).each do |key, value|
+        ENV[key.to_s] = value
+      end if File.exists?(env_file)
+    end
```
zediogoviana, line 52: Are you sure this is needed? Also, why are you using a YAML file to set env vars?

config/environments/development.rb
RicardoVaz1: I was just testing if the Action Mailer was actually sending emails, I'll remove that and just leave the Mailcatcher

config/application.rb
RicardoVaz1: I saw [this](https://railsapps.github.io/rails-environment-variables.html) solution for the ENV variables, "Option Three: Use a local_env.yml File", and it says that we need to have that because "The code opens the config/local_env.yml file, reads each key/value pair, and sets environment variables."

zediogoviana: I don't think you need a gem or even this... Wasn't this working previously? In the other PR you already accessed env vars configured in the `.env` file, right?

db/seeds.rb
```diff
@@ -14,5 +14,6 @@
   Prize.create!(name: "NFT #{prize + 1}", percentage: 10, image: "assets/svgs/prizes/nft.svg", prize_type: 'NFT', description: 'This is a NFT item', identifier: 'NFT')
 end
 
-Lead.create(name: "test", email: "test@subvisual.com", interests: ["Development"], data_proc_consent: true)
-AdminUser.create!(email: 'admin@example.com', password: 'password', password_confirmation: 'password') if Rails.env.development?
\ No newline at end of file
+
```
zediogoviana, line 17: Why the extra line here if you already have one above? One empty line is enough 😅

config/application.rb
RicardoVaz1: Yes, because I had that code. And I ran the Backend first and then I went to the console, so that's why I had the results from the ENV's file. If I comment that code and go to the console I'll get this:

print_

db/seeds.rb
RicardoVaz1: You're right, I'll fix it

config/application.rb
zediogoviana: Have you sourced the file?

RicardoVaz1: Probably not, how do I do that?

zediogoviana: Without the server or console running (or if they are running you need to restart later) and do `source .env.dev` or whatever the env file you have (not the yaml one).

DavideSilva: Or you could use something like [`direnv`](https://direnv.net/) to automatically set the environment variables in that folder

RicardoVaz1: I commented out that yaml code and created the files .env and .env.dev  (inside the config folder) and I ran `source .env` and `source .env.dev`, and I also tried in the console but didn't work (print 1). So I tried the same ideia with those files in the root folder but also didn't work (print 2). 

Screenshot from 2022-08-12 10-20-50

Screenshot from 2022-08-12 10-22-39

I found [this](https://github.com/bkeepers/dotenv) information about ENV variables, but it requires installing a gem. Is this the right way to setup the ENV variables?

DavideSilva: How did you setup variables in the `.env` or `.env.dev`?  You probably need to add an `export` before the environment variable name, like `export NAME="name"`

RicardoVaz1: I just noticed your idea about the `direnv` after sending the message, but I'll look into it

RicardoVaz1: I forgot the export, I will add the export and try again

RicardoVaz1: Adding the export worked on the console, but when I run the app to test it, it's still not working. Do I need to add something like `require("dotenv").config()` at the top of the files where I want to use the ENV variables?

zediogoviana (approved):

RicardoVaz1:
Adding the export worked in the console, but when I test the app it gives me this error: `KeyError (key not found: "SERVER_URL"):`. I was looking at [this](https://www.rubydoc.info/gems/dotenv-rails/2.1.1#sinatra-or-plain-ol-ruby) link, will it be necessary to install a gem and then use:
```
require 'dotenv'
dotenv.load
```
in the files where I am using ENV.fetch('SERVER_URL') ?

config/application.rb
DavideSilva: I don't think you need to have the `dotenv` gem, Rails usually takes care of loading environment variables automatically. In fact, the REACT_APP_BACKEND_PATH variable is already being loaded

zediogoviana (approved):

PR #101, opened 2022-08-11
Add back button component

jreis13:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5216095002
* We want to give the user the opportunity to go back in the signup screens and policy page.

How:
* Add the `<BackButton />` component and necessary styles.

### DIFF
--- a/frontend/src/assets/svgs/wheel-background/chevron.svg
+++ b/frontend/src/assets/svgs/wheel-background/chevron.svg
@@ -1,0 +1,3 @@
+<svg fill="currentColor" xmlns="http://www.w3.org/2000/svg">
+  <path d="M23.347 5.16 20.973 2.8 7.787 16l13.2 13.2 2.36-2.36L12.507 16l10.84-10.84Z" />
+</svg>
--- a/frontend/src/components/BackLink/index.js
+++ b/frontend/src/components/BackLink/index.js
@@ -1,0 +1,20 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.scss";
+
+import { ReactComponent as Chevron } from "assets/svgs/wheel-background/chevron.svg";
+
+const BackLink = ({ href }) => {
+  return (
+    <div className={styles.backLinkWrapper}>
+      <a href={href} className={styles.backLink}>
+        <Chevron />
+      </a>
+    </div>
+  );
+};
+
+BackLink.propTypes = {
+  href: PropTypes.string.isRequired,
+};
+
+export default BackLink;
--- a/frontend/src/components/BackLink/index.module.scss
+++ b/frontend/src/components/BackLink/index.module.scss
@@ -1,0 +1,14 @@
+.backLinkWrapper {
+  margin: 24px 0 92px 24px;
+}
+
+.backLink {
+  width: 32px;
+  height: 32px;
+  color: var(--color-light-pink);
+
+  > svg {
+    width: inherit;
+    height: inherit;
+  }
+}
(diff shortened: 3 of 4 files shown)


frontend/src/assets/svgs/wheel-background/back-button.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="32" height="32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mirhamasala, line 1: Same story:

- Needs to be run through the optimizer
- Remove `width`, `height`
- Set `fill` to `currentColor`
- Add `color` in component

frontend/src/components/BackButton/index.js
```diff
@@ -0,0 +1,18 @@
...
+const BackButton = ({ href }) => {
+  return (
+    <a href={href} className={styles.backButton}>
+      <img src={BackButtonLogo} alt="Back button"></img>
+    </a>
+  );
+};
+
+BackButton.propTypes = {
+  href: PropTypes.string,
```
mirhamasala, line 15: I'd say this is `required`

frontend/src/components/BackButton/index.js
```diff
@@ -0,0 +1,18 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.css";
+
+import BackButtonLogo from "assets/svgs/wheel-background/back-button.svg";
+
+const BackButton = ({ href }) => {
+  return (
+    <a href={href} className={styles.backButton}>
+      <img src={BackButtonLogo} alt="Back button"></img>
```
mirhamasala, line 9: Since this is an SVG in React, let's import them directly as components.

https://create-react-app.dev/docs/adding-images-fonts-and-files#adding-svgs

frontend/src/components/BackButton/index.module.css
```diff
@@ -0,0 +1,4 @@
+.backButton {
+  width: fit-content;
+  margin: 24px 0 92px 24px;
```
mirhamasala, line 3: It's not good practice to add layout styles to the component directly since you now can't reuse this `<` as is anywhere else.

I'm okay to leave it in for now, but you can imagine that you would want to extract something like this into a `header` or the likes.

frontend/src/components/BackButton/index.module.css
```diff
@@ -0,0 +1,4 @@
+.backButton {
+  width: fit-content;
```
mirhamasala, line 2: What do you think this does?

frontend/src/components/BackButton/index.module.css
```diff
@@ -0,0 +1,4 @@
+.backButton {
```
mirhamasala, line 1:
```suggestion
.root {
```

frontend/src/components/BackButton/index.js
```diff
@@ -0,0 +1,18 @@
+import { PropTypes } from "prop-types";
+import styles from "./index.module.css";
+
+import BackButtonLogo from "assets/svgs/wheel-background/back-button.svg";
+
+const BackButton = ({ href }) => {
```
mirhamasala, line 6: I'd say this is a `BackLink` not a `Button`

frontend/src/components/BackButton/index.module.css
mirhamasala: Or `.link`

frontend/src/components/BackButton/index.module.css
mirhamasala: Or have a `wrapper` class in the parent like you do for all the other components.

frontend/src/components/BackButton/index.module.css
mirhamasala: Let's add `width` and `height` values directly and let the SVG inherit them.

mirhamasala (approved):
Just 👌🏼👌🏽👌🏾! LGTM.

PR #102, opened 2022-08-11
Add max probability to prizes

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5184186427

How:
* Creating a function to sum all the prizes percentages
* Adding a validation on the prize model

### DIFF
--- a/app/controllers/prizes_controller.rb
+++ b/app/controllers/prizes_controller.rb
@@ -7,8 +7,13 @@ class PrizesController < ApiController
   end
 
   def create
-    prize = Prize.create(prize_params)
-    render json: prize
+    total_prizes_percentages = Prize.pluck(:percentage).sum
+    if total_prizes_percentages > 100 || total_prizes_percentages + params[:percentage] > 100
+      render json: 'Error! The total percentage is greater than 100%!'
+    else
+      prize = Prize.create(prize_params)
+      render json: prize
+    end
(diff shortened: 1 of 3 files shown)


app/controllers/prizes_controller.rb
```diff
@@ -25,6 +29,16 @@ def win_prize
...
+  def sum_prizes_percentages
+    total_percentage = 0
+
+    Prize.all.each_with_object({}) do |prize, _acc|
+      total_percentage += prize.percentage
+    end
+
+    total_percentage
+  end
+
```
DavideSilva, line 41: There's a one line solution for this:

```
Prize.pluck(:percentage).sum
```

`pluck` is a method that retrieves all the values from the column/attribute we pass as parameter.
`Prize.pluck(:percentage)` will retrieve an Array with the percentage values for all the Prizes. Since we are simply adding all the percentages, we can call the `sum` method of the Array class to get the final sum

mirhamasala: Nothing like a one-liner to replace 10 lines. So fresh and so clean clean. 🧼

RicardoVaz1: Thanks, it works!

DavideSilva (approved):
:shipit:

PR #110, opened 2022-08-12
Remove the need to select one of the interests

jreis13:
Why:
* After meeting with the product owner it was decided that it should be optional to pick an interest.

How:
* Removed the `else if` that was blocking users from moving to the next page if they hadn't selected any interest.


mirhamasala (approved):
💙🚀✨

PR #112, opened 2022-08-12
Converting policy page into a policy form inside the sign up page

beasrodrigues24:
Why:
* For it to be possible to use a button to return to a previous step, the 4 steps must be in the same page

How:
* By moving PolicyPage files to components/Forms/PolicyForm
* By creating a new step on the sign up and adapting across the code to fit this


frontend/src/pages/SignupPage/index.js
```diff
@@ -141,9 +175,15 @@ const SignupPage = () => {
         <Logo />
       </div>
       <div className={styles.main}>{stepInfo[state.step].component}</div>
-      <Button onClick={continueToNextStep}>
-        {stepInfo[state.step].buttonText}
-      </Button>
+      {(() => {
+        if (state.step !== STEP_4) {
```
beasrodrigues24, line 179: This if exists because the button on the step 4 will be styled differently with the new design

beasrodrigues24: So it will come from inside the PrivacyForm component

beasrodrigues24:
Also the Privacy page will be a little crooked, it will have to be fixed in another PR after we decide if we are keeping the Logo on top of the page or not

frontend/src/pages/SignupPage/index.js
mirhamasala: Why the function?

mirhamasala:
Can you please check the CSS for this form? It overlaps with the SignUp page. Especially, `.root` and `.main` You might not need that.

frontend/src/pages/SignupPage/index.js
mirhamasala: Is this what you're looking for?

```
      {state.step !== STEP_4 && (
        <Button onClick={continueToNextStep}>
          {stepInfo[state.step].buttonText}
        </Button>
      )}
```

frontend/src/components/Button/index.js
```diff
@@ -1,21 +1,33 @@
...
-    <button onClick={onClick} className={styles.button}>
+    <button onClick={onClick} className={buttonClassName}>
       {children}
     </button>
   );
 };
 
+Button.defaultProps = {
+  isDisabled: false,
+  href: "",
```
mirhamasala, line 24: Why set this to an empty string?

frontend/src/pages/SignupPage/index.js
```diff
@@ -158,27 +158,22 @@ const SignupPage = () => {
...
     },
     STEP_4: {
+      buttonText: "I agree",
       component: (
         <PolicyForm handleChange={handleChange} submitData={submitData} />
       ),
     },
   };
 
+  let buttonHandler = state.step === STEP_4 ? submitData : continueToNextStep;
```
mirhamasala, line 168: Why not `const`?

beasrodrigues24: Oh right, João explained me better the difference between the two. I'll change it!

mirhamasala (approved):
✨❤️ Awesome. LGTM! 🚀

PR #113, opened 2022-08-12
Add email verification

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5223467271

How:
* Using a regex expression


joaojusto (changes requested):
Email validation is considered a bad practice. For these purposes, we should only check if the email has something in it or if it's empty.

https://www.loqate.com/resources/blog/3-reasons-why-you-should-stop-using-regex-email-validation/

RicardoVaz1:
Ok I didn't know that. Should I delete this PR then?

joaojusto (approved):

PR #115, opened 2022-08-12
Implementing backlink and refactoring gradient container

beasrodrigues24:
Why:
* Should be possible to turn back on sign up form steps
* ContainerWithGradient should be possible to use without a gradient

How:
* By using a back link component
* By refactoring ContainerWithGradient to MainContainer


frontend/src/components/BackLink/index.module.scss
```diff
@@ -1,5 +1,13 @@
 .backLinkWrapper {
-  margin: 24px 0 92px 24px;
+  margin: 24px 0 24px 24px;
+}
+
+button {
```
mirhamasala, line 5: Did the linter not complain about using `button` in the wild? Without converting it to a class?

frontend/src/components/BackLink/index.js
```diff
@@ -3,18 +3,18 @@ import styles from "./index.module.scss";
 
 import { ReactComponent as Chevron } from "assets/svgs/wheel-background/chevron.svg";
 
-const BackLink = ({ href }) => {
+const BackLink = ({ handleReturn }) => {
   return (
     <div className={styles.backLinkWrapper}>
-      <a href={href} className={styles.backLink}>
+      <button onClick={handleReturn} className={styles.backLink}>
```
mirhamasala, line 9: Oe. It's late and I'm blanking. But `handleReturn` reminds me too much of `return` ....

frontend/src/components/MainContainer/index.js
```diff
@@ -0,0 +1,23 @@
+import { PropTypes } from "prop-types";
```
mirhamasala, line 1: Excellent! You're picking up so fast!

frontend/src/components/MainContainer/index.module.scss
```diff
@@ -2,5 +2,9 @@
   display: flex;
   flex-direction: column;
   min-height: 100vh;
-  background: linear-gradient(var(--color-light-pink), var(--color-white));
+  background-color: var(--color-white);
 }
+
+.withGradient {
```
mirhamasala, line 8: `.hasGradient` to be consistent?

frontend/src/pages/SignupPage/index.js
```diff
@@ -107,6 +125,11 @@ const SignupPage = () => {
       });
   };
 
+  const continueToPrevStep = () => {
```
mirhamasala, line 128: `goBackToPreviousStep` ?

frontend/src/pages/SignupPage/index.module.css
```diff
@@ -1,21 +1,30 @@
...
 }
 
 .logo {
   display: flex;
   justify-content: flex-end;
-  padding: 50px;
+  margin-bottom: 45px;
+}
+
+.backLinkInvisible {
```
mirhamasala, line 21: `.isHidden` ?

frontend/src/pages/SignupPage/index.module.css
```diff
@@ -1,21 +1,30 @@
...
+  visibility: hidden;
+}
+
+.buttonWrapper {
+  position: relative;
```
mirhamasala, line 26: Why?

frontend/src/pages/SignupPage/index.js
```diff
@@ -168,13 +191,26 @@ const SignupPage = () => {
...
-      <Button onClick={buttonHandler}>{stepInfo[state.step].buttonText}</Button>
-    </div>
+    <>
+      <MainContainer>
+        <div
+          className={state.step === STEP_1 ? styles.backLinkInvisible : null}
+        >
+          <BackLink handleReturn={continueToPrevStep} />
+        </div>
+        <div className={styles.root}>
```
mirhamasala, line 201: It's not a `root` anymore if it appears ten lines down. 🙃

frontend/src/pages/SignupPage/index.module.css
```diff
@@ -1,21 +1,30 @@
...
+  display: flex;
+  justify-content: center;
+  margin-bottom: 60px;
```
mirhamasala, line 29: What about adding `padding-bottom: 60px` to the `MainContainer`?

If that works, you need to make sure to remove the spacing from other individual components.

frontend/src/pages/SignupPage/index.js
```diff
@@ -168,13 +191,26 @@ const SignupPage = () => {
...
-    <div className={styles.root}>
-      <div className={styles.logo}>
-        <Logo />
-      </div>
-      <div className={styles.main}>{stepInfo[state.step].component}</div>
-      <Button onClick={buttonHandler}>{stepInfo[state.step].buttonText}</Button>
-    </div>
+    <>
+      <MainContainer>
+        <div
```
mirhamasala, line 196: So instead of having a blank `div` here, you can now move the `BackLinkWrapper` styles from the `BackLink` component into here.

I mentioned this previously to @jreis13 so he can maybe explain it if necessary. Basically, I explained to him that layout styles shouldn't be applied to a component like `BackLink` but to the parent.

That way, this `div` would be `BackLinkWrapper` and you'd chain a class `.isHidden` when needed.

frontend/src/pages/SignupPage/index.module.css
mirhamasala: I think you can remove `buttonWrapper` altogether.

frontend/src/pages/SignupPage/index.js
```diff
@@ -168,13 +191,26 @@ const SignupPage = () => {
...
 
   return (
-    <div className={styles.root}>
-      <div className={styles.logo}>
-        <Logo />
-      </div>
-      <div className={styles.main}>{stepInfo[state.step].component}</div>
-      <Button onClick={buttonHandler}>{stepInfo[state.step].buttonText}</Button>
-    </div>
+    <>
```
mirhamasala, line 194: When you have a component like `MainContainer` which encapsulates all the other elements, you don't need a fragment anymore `<>`

frontend/src/components/BackLink/index.module.scss
```diff
@@ -1,5 +1,13 @@
 .backLinkWrapper {
```
mirhamasala, line 1: If it's not a `backLink` anymore but a `button`, the name should be changed.

frontend/src/components/BackLink/index.module.scss
mirhamasala: The property order is also whack according to linter styles. I think your linter isn't working?

mirhamasala (changes requested):
I realize I've pointed out quite a few things, but they're details. You picked up really fast and did a great job! (Both of you? Also @jreis13)  ✨

frontend/src/components/BackLink/index.module.scss
jreis13: I can run this branch through my VSCode tomorrow and maybe it will fix this issue

frontend/src/pages/SignupPage/index.module.css
beasrodrigues24: We tried to remove it but it didn't work. So @jreis13 put display and justify-content only and that works!

frontend/src/components/BackLink/index.module.scss
```diff
@@ -1,11 +1,15 @@
...
 
-.backLink {
+.button {
   width: 32px;
   height: 32px;
+  padding: 0;
   color: var(--color-light-pink);
+  background: none;
+  border: none;
+  outline: inherit;
```
joaojusto, line 12: Where is the backlink inheriting the outline from? 🤔

frontend/src/pages/SignupPage/index.module.css
```diff
@@ -1,21 +1,28 @@
...
 
 .logo {
   display: flex;
   justify-content: flex-end;
-  padding: 50px;
+  margin-bottom: 45px;
+}
+
+.isHidden {
+  visibility: hidden;
```
joaojusto, line 22: Do we want to just toggle the visibility here? I mean, do we want to just hide it but keep it in the DOM, accessible for screen readers and taking space in the document, or do we want to remove it completely?

jreis13: I think we want to just hide it and 'reserve' that space

frontend/src/components/BackLink/index.module.scss
jreis13: Was a mistake leftover, just wanted to remove the default button styles. going to fix that.

frontend/src/pages/SignupPage/index.module.css
joaojusto: Cool then it's good to go 👌

joaojusto (approved):
LGTM, after fixing the `.button`

PR #119, opened 2022-08-16
Adding redirection from route /admin to admin backoffice

beasrodrigues24:
Why:
* It allows us to not have to memorize another link

How:
* By creating a /admin route and an admin page that does the redirection

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -5,6 +5,7 @@ import PrizePage from "pages/PrizePage";
 import SignupPage from "pages/SignupPage";
 import SocialPage from "pages/SocialPage";
 import WheelPage from "pages/WheelPage";
+import AdminPage from "pages/AdminPage";
 
 function App() {
   return (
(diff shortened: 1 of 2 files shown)


joaojusto (changes requested):
I must say that I love the hack but I do think you should do the redirect using React router. I think instead of a custom `AdminPage` as a route you could use as element `<Navigate/>`. Take a look at [`Navigate`](https://reactrouter.com/docs/en/v6/components/navigate)

beasrodrigues24:
I tried navigate but didn't find a way to make it go to an absolute path, so I went through this route. But I'll look further into it to find how it's done!

joaojusto:
@beasrodrigues24 fair enough, maybe you're right. Did you tried with `replace`?  `<Navigate replace to="https://subvisuwheel.herokuapp.com/admin/login" />`?

beasrodrigues24:
Sadly, same issues 
image

joaojusto (approved):
@beasrodrigues24 then let's merge as it is

PR #120, opened 2022-08-16
Style policy page

jreis13:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5212082783
* Need to accept the terms and conditions to move forward
* We should not use a link directly on the label so we had to change the `CheckBox` component's behaviour by wrapping the `label` and an additional `children` prop in a `div` so that it still looks like the same component when we use the `<a>` tag with the link to the terms and conditions external page.

How:
* Made changes in the `Checkbox` component to be able to implement logic and style.
* Changed the `PolicyForm` component 
* Added disabled button on start that activates once we accept the t&c


joaojusto:
@jreis13 I would love to challenge you to expand on the `Why` we needed to change the `<Checkbox/>`. Talking about how the parent is using `flex` and why that wasn't the positioning behavior we wanted.

frontend/src/components/Checkbox/index.module.css
```diff
@@ -1,26 +1,30 @@
...
+  margin-right: 20px;
+}
+
 .input {
+  position: absolute;
+  left: 0;
   width: var(--size-checkbox);
   height: var(--size-checkbox);
   opacity: 0;
 }
```
joaojusto, line 20: Do we need any of this? Couldn't we achieve the same functionality and design with: 
```.css
.input {
  position: absolute;
  visibility: hidden;
}
```
?

frontend/src/components/Checkbox/index.module.css
```diff
@@ -1,26 +1,30 @@
...
+.root {
   --size-checkbox: 24px;
 
   position: relative;
   display: flex;
   align-items: center;
 }
 
+.inputWrapper {
+  width: var(--size-checkbox);
```
joaojusto, line 10: I don't think we need to explicitly set the `width` anymore since `.control` is no longer absolutely positioned. Am I wrong?

frontend/src/components/Forms/PolicyForm/index.module.scss
```diff
@@ -0,0 +1,19 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  padding: 0 20px;
+}
+
+.termsAndConditions {
+  margin-bottom: 32px;
+
+  a {
```
joaojusto, line 16: Does the link have padding by default? 🤔

joaojusto: If we answer is yes, shouldn't we move this to `normalize`?

joaojusto (changes requested):

frontend/src/components/Checkbox/index.module.css
jreis13: by defining visibility to be hidden the ☑️ doesn't show at all. 

What does work is:
```
.input {
  position: absolute;
  opacity: 0;
}
```

frontend/src/components/Forms/PolicyForm/index.module.scss
jreis13: It didn't, was just a style I forgot to erase while testing things out - it has been removed already

joaojusto (approved):

PR #121, opened 2022-08-16
Fix bug of not saving checkbox status

beasrodrigues24:
Why:
* If someone checked one or more boxes and turned back a step and returned the boxes wouldn't be checked. This would induce the user to select again, however this would delete the selection altogether and make it as the user didn't select anything.

How:
* By adding a checked attribute to the checkbox
* By having a validation function that checks if a box has been checked

### DIFF
--- a/frontend/src/pages/SignupPage/index.js
+++ b/frontend/src/pages/SignupPage/index.js
@@ -80,6 +80,18 @@ const SignupPage = () => {
       }
     };
 
+  const validatesChecked = (id) => {
+    if (id === "other") {
+      return state.otherChecked;
+    } else if (id === "jobs") {
+      return state.consent;
+    } else if (id === "terms") {
+      return state.termsChecked;
+    } else {
+      return state.selectedInterests.includes(id);
+    }
+  };
+
   const prevStep = () => {
     const { step } = state;
     const newStep = () => {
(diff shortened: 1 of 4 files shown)


beasrodrigues24:
This PR is on hold until João has another PR go through because I'll probably have to fix his bus in there as well!

frontend/src/pages/SignupPage/index.js
```diff
@@ -80,6 +80,18 @@ const SignupPage = () => {
       }
     };
 
+  const validatesChecked = (id) => {
+    if (id === "other") {
+      return state.otherChecked;
+    } else if (id === "jobs") {
+      return state.consent;
+    } else if (id === "terms") {
```
joaojusto, line 88: What about extracting these to constants?

joaojusto (changes requested):
Just a minor comment but still would be cool to fix it. You can export the constants you're going to be create and import them where you cal the validation function

frontend/src/pages/SignupPage/index.js
beasrodrigues24: Okay seems good, will do that in the afternoon after I finish my current task :smile:

joaojusto (approved):

PR #124, opened 2022-08-16
Fix size of input hidden behind the custom checkbox

beasrodrigues24:
Why:
* It was too small so it didn't overlap the whole box and it was possible to click in most of the box without having any effect
image 
(Colors don't reflect the actual colors since I'm using dark mode)

How:
* By setting width and height for the input that match the control box

### DIFF
--- a/frontend/src/components/Checkbox/index.module.css
+++ b/frontend/src/components/Checkbox/index.module.css
@@ -12,6 +12,8 @@
 
 .input {
   position: absolute;
+  width: var(--size-checkbox);
+  height: var(--size-checkbox);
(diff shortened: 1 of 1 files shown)


joaojusto (changes requested):
@beasrodrigues24 we might need to center the input vertically in the case where the label has 2 lines.

joaojusto (approved):

beasrodrigues24:
> @beasrodrigues24 we might need to center the input vertically in the case where the label has 2 lines.

I don't really understand what changes I'd have to make for this

PR #125, opened 2022-08-16
Adapt extra interest to be discarded if other is unchecked

beasrodrigues24:
Why:
* If a user clicked on other, wrote something and then unchecked, we would still keep the value.

How:
* By introducing a validation
* By removing if (STEP_4) that wasn't being executed
* By moving the validation to the data submition so that the user has until that moment to decide if he does want to keep the "Other"


beasrodrigues24:
I have to rebase this, has commits from other branches!

joaojusto (approved):

PR #127, opened 2022-08-17
Add database for the company information

beasrodrigues24:
Why:
* It should be in the database for it to be added to active admin in a further step

How:
* By creating a Company and Social model
* By creating a Company controller that collects the socials as well and does serialization
* By creating a route that allows a get request to get the socials
* By ensuring that there only exists one selected company at a time

image

### DIFF
--- a/app/controllers/socials_controller.rb
+++ b/app/controllers/socials_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# Controller that manages socials
+class SocialsController < ApiController
+  def index
+    render json: Social.all
+  end
+
+  def show
+    render json: serialize_socials(Company.selected)
+  end
+
+  private
+
+  def serialize_socials(company)
+    company.socials.each_with_object({}) do |social, acc|
+      acc[social.name.downcase] = social.url
+    end
+  end
+end
--- a/app/models/company.rb
+++ b/app/models/company.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# Model that allows storing companies
+class Company < ApplicationRecord
+  before_save :ensure_only_one_selected
+
+  validates :name, :website, :selected, presence: true
+
+  has_many :social
+
+  private
+
+  def ensure_only_one_selected
+    self.class.update_all(selected: false)
+  end
+end
--- a/app/models/company.rb
+++ b/app/models/company.rb
@@ -8,6 +8,8 @@ class Company < ApplicationRecord
 
   has_many :social
 
+  scope :is_selected, -> { find_by(selected: true) }
+
   private
 
   def ensure_only_one_selected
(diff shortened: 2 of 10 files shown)


beasrodrigues24:
Decided to split my task in 3 PR's. The next 2 will be adding this to active admin and changing the frontend.

app/models/company.rb
```diff
@@ -0,0 +1,18 @@
...
+
+# Model that allows storing companies
+class Company < ApplicationRecord
+  before_save :ensure_only_one_selected
+
+  validates :name, :website, :selected, presence: true
+
+  has_many :social
+
+  scope :is_selected, -> { find_by(selected: true) }
```
zamith, line 11:
```suggestion
  scope :selected, -> { find_by(selected: true) }
```

Rubyist don't use `is` a lot.

app/models/company.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Model that allows storing companies
+class Company < ApplicationRecord
+  before_save :ensure_only_one_selected
+
+  validates :name, :website, :selected, presence: true
+
+  has_many :social
```
zamith, line 9:
```suggestion
  has_many :socials
```

If it's many, it makes sense to be plural, no?

beasrodrigues24:
So after seeing the new designs, it wasn't using the company name or website so I adapted it to only send to the frontend the socials part but keep the possibility to access everything because it'll probably be necessary for the emails

config/routes.rb
```diff
@@ -5,6 +5,7 @@
   ActiveAdmin.routes(self)
   get '/prizes', to: 'prizes#index'
   post '/prizes', to: 'prizes#create'
+  get '/socials', to: 'companies#socials_from_company'
```
zamith, line 8: Not a big fan of this, tbh. As a rule of thumb, especially as you're getting started, try to only use RESTful actions. For example, instead of `/socials` going to the `companies` controller you can do `/companies/:id/socials`, or in your case, since you're using the `selected` attribute, you can do `/socials` map to `socials#show` or `selected_company_socials#show`. Don't be afraid to have many controllers.

zamith (changes requested):

app/controllers/socials_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Controller that manages socials
```
zamith, line 3: I would remove this comment. In general I'm not a fan of comments unless they really add something. Most times you can get rid of them with clearer naming. This one doesn't explain anything new, imo. What do you think?

zamith (approved):

app/controllers/socials_controller.rb
beasrodrigues24: I didn't want it there but rubocop forced me :sweat_smile:

PR #128, opened 2022-08-17
Add company and socials to backoffice

beasrodrigues24:
Why:
* For it to be possible to add companies, select the company using the wheel and link socials to the companies

How:
* By generating the admin resources

### DIFF
--- a/app/admin/companies.rb
+++ b/app/admin/companies.rb
@@ -1,18 +1,5 @@
 # frozen_string_literal: true
 
 ActiveAdmin.register Company do
-  # See permitted parameters documentation:
-  # https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
-  #
-  # Uncomment all parameters which should be permitted for assignment
-  #
   permit_params :name, :website, :selected
-  #
-  # or
-  #
-  # permit_params do
-  #   permitted = [:name, :website, :selected]
-  #   permitted << :other if params[:action] == 'create' && current_user.admin?
-  #   permitted
-  # end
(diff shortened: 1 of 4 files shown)


app/admin/companies.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register Company do
+  # See permitted parameters documentation:
```
zamith, line 4: Do you need these comments?

app/controllers/companies_controller.rb
```diff
@@ -7,8 +7,8 @@ def index
   end
 
   def serialize_company(company)
-    serialized_socials = Social.where(company_id: company.id).map do |social|
-      social.slice(:name, :url)
+    serialized_socials = Social.where(company_id: company.id).all.each_with_object({}) do |social, acc|
```
zamith, line 10:
```suggestion
    serialized_socials = company.socials.each_with_object({}) do |social, acc|
```

What about this?

beasrodrigues24: Will fix in the other PR and rebase :smile:

app/admin/companies.rb
beasrodrigues24: What comment to you mean? The documentation or the frozen_string_literal? If it's the second rubocop forces me to have that. If it's the documentation I can remove it

zamith: The documentation. I understand it's auto generated, but if you don't need it, remove it.

zamith (approved):
Would remove the comments, but it's not a blocker

PR #129, opened 2022-08-17
Adapt sign up to design

jreis13:
Why:
* Sign-up flow was not updated with the new design.

How:
* Removed `Logo` from most pages since it is not present in the design;
* Fixed copy to match Figma;
* Fixed some minor details in the privacy policy section;
* Added the disabled state to the `button` when no `input` is provided in the first and second steps of the sign-up;
* Moved the side `padding` to the `MainContainer` and removed it on the pages that are using `.hasGradient`.


joaojusto (approved):
LGTM

PR #131, opened 2022-08-17
Configure Emails in Admin Backoffice

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5227038325

How:
* Creating a Email controller in Active Admin

### DIFF
--- a/app/admin/emails.rb
+++ b/app/admin/emails.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+ActiveAdmin.register Email do
+  # See permitted parameters documentation:
+  # https://github.com/activeadmin/activeadmin/blob/master/docs/2-resource-customization.md#setting-up-strong-parameters
+  #
+  # Uncomment all parameters which should be permitted for assignment
+  #
+
+  permit_params :email_type, :subject, :body
+
+  #
+  # or
+  #
+  # permit_params do
+  #   permitted = [:email_type, :subject, :body]
+  #   permitted << :other if params[:action] == 'create' && current_user.admin?
+  #   permitted
+  # end
+end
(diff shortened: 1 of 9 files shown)


app/controllers/emails_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+# Controller that manages emails
+class EmailsController < ApiController
+  def index
+    render json: Email.all
+  end
+
+  def create
+    email = Email.create(email_params)
+    render json: email
+  end
```
zediogoviana, line 12: If you are moving this to be configurable in the backend, why doesn't it also exist an update? I assume that there is a good reason for this to be configurable this way, and not just have them in the code...

app/admin/emails.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  #
+  # or
+  #
+  # permit_params do
+  #   permitted = [:email_type, :subject, :body]
+  #   permitted << :other if params[:action] == 'create' && current_user.admin?
+  #   permitted
+  # end
+end
```
zediogoviana, line 20: Are all of these comments supposed to be here? The same thing for the other files that have comments. I don't think this adds any valuable information.

app/controllers/emails_controller.rb
RicardoVaz1: The Active Admin has his own controller that have already an update function. I'm just using that create function to create the default templates in the seeds file

EmailsAdminBackffice

zediogoviana: Oh, right! I remember now 😅

app/admin/emails.rb
RicardoVaz1: No, I'll delete them

zediogoviana (approved):
👌

PR #137, opened 2022-08-18
Close random access to /wheel and /win-prize

beasrodrigues24:
Why: 
* The user shouldn't be able to access these pages if he doesn't do the flow correctly because he'll have blank error screens

How:
* Redirecting the user back to the home screen with useEffect hook that validates if the local storage suggests the user has played
* Doing this in two components, one for the Wheel (checks if the email exists) and one for the prize (checks if the user has won a prize)

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -1,5 +1,8 @@
 import { Routes, Route } from "react-router-dom";
 
+import ValidateEmail from "components/ValidateEmail";
+import ValidatePrize from "components/ValidatePrize";
+
 import HomePage from "pages/HomePage";
 import PrizePage from "pages/PrizePage";
 import SignupPage from "pages/SignupPage";
@@ -13,9 +16,23 @@ function App() {
       <Routes>
         <Route path="/" element={<HomePage />} />
         <Route path="/signup" element={<SignupPage />} />
-        <Route path="/wheel" element={<WheelPage />} />
+        <Route
+          path="/wheel"
+          element={
+            <ValidateEmail>
+              <WheelPage />
+            </ValidateEmail>
+          }
+        />
         <Route path="/social-network" element={<SocialPage />} />
-        <Route path="/win-prize" element={<PrizePage />} />
+        <Route
+          path="/win-prize"
+          element={
+            <ValidatePrize>
+              <PrizePage />
+            </ValidatePrize>
+          }
+        />
         <Route path="/admin" element={<AdminPage />} />
       </Routes>
     </>
--- a/frontend/src/components/ValidateEmail/index.js
+++ b/frontend/src/components/ValidateEmail/index.js
@@ -1,0 +1,24 @@
+import { PropTypes } from "prop-types";
+import { useNavigate } from "react-router-dom";
+import { useEffect, useState } from "react";
+
+const ValidatePrize = ({ children }) => {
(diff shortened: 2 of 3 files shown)


joaojusto (changes requested):
@beasrodrigues24 can we extract this logic to an higher-order component/s? Something like: 
```
<Route path="/win-prize" element={
  <RequirePrize>
    <PrizePage />
  </RequirePrize>
} />
```
```
<Route path="/wheel" element={
  <RequireEmail>
    <WheelPage />
  </RequireEmail>
} />
```

What do you think?

beasrodrigues24:
> @beasrodrigues24 can we extract this logic to an higher-order component/s? Something like:
> 
> ```
> <Route path="/win-prize" element={
>   <RequirePrize>
>     <PrizePage />
>   </RequirePrize>
> } />
> ```
> 
> ```
> <Route path="/wheel" element={
>   <RequireEmail>
>     <WheelPage />
>   </RequireEmail>
> } />
> ```
> 
> What do you think?

On It!

beasrodrigues24:
Will delete the first 2 commits after this PR is done. Will let them be for reference only for now

frontend/src/components/RequireValidation/index.js
```diff
@@ -0,0 +1,29 @@
...
+import { useNavigate } from "react-router-dom";
+import { useEffect, useState } from "react";
+
+const RequireValidation = ({ children }) => {
+  const navigator = useNavigate();
+  const [render, setRender] = useState(null);
+
+  useEffect(() => {
+    if (
+      (children.type.componentName === "WheelPage" &&
```
joaojusto, line 11: If we change the component names the validation breaks. I don't think that makes sense,  why can't we have separated components for each type of validation?

frontend/src/components/RequireValidation/index.js
```diff
@@ -0,0 +1,29 @@
...
+      (children.type.componentName === "PrizePage" &&
+        localStorage.getItem("prizeWon"))
+    ) {
+      setRender(render);
+    } else {
+      navigator("/");
+    }
+  }, [children.type.componentName, render, navigator]);
+
+  return render;
```
joaojusto, line 22: Why can't we just return `children` here?

joaojusto (changes requested):

frontend/src/components/RequireValidation/index.js
beasrodrigues24: I tried and it would skip the useEffect and try to render the page, so I had to find a work around

frontend/src/components/RequireValidation/index.js
beasrodrigues24: Okay!

frontend/src/components/RequireValidation/index.js
joaojusto: I would instead set a loading flag, and either return children or null

beasrodrigues24: On it!

beasrodrigues24:
Don't review yet, i messed up rebasing this !

joaojusto (approved):
Much better 🚢

joaojusto (commented):
Small, focused, and reusable components for the win

PR #138, opened 2022-08-18
Fix signup page

jreis13:
Why:
* User interviews showed that users prefer having the name and email inputs in the same page

How:
* Refactor signup flow to match the design (moved the name form and email form into a single `DetailsForm`)
* Input boxes now have labels to match the design
* Update the `placeholders` on the inputs
* Remove `STEP_4` as we only have 3 steps now
* Add constraint so that the user has to type an '@' so that we let him move forward past the email input


frontend/src/components/Forms/FormLayout/index.module.css
```diff
@@ -1,5 +1,5 @@
 .headerWrapper {
-  padding: 0 20px;
+  /* padding: 0 20px; */
```
joaojusto, line 2: If this is not needed please remove it

frontend/src/pages/SignupPage/index.js
```diff
@@ -244,9 +229,10 @@ const SignupPage = () => {
       <div className={styles.buttonWrapper}>
         <Button
           isDisabled={
```
joaojusto, line 231: Can we extract this into a variable? It's a bit verbose to be inline

joaojusto (changes requested):
Looks good, just minor details

frontend/src/components/Forms/FormLayout/index.module.css
jreis13: You're right, on it.

joaojusto (approved):
:shipit:

PR #141, opened 2022-08-18
Fix Prizes validation is above 100%

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5238453637

How:
* Creating a validation in the Admin Prizes controller

### DIFF
--- a/app/admin/prizes.rb
+++ b/app/admin/prizes.rb
@@ -2,4 +2,33 @@
 
 ActiveAdmin.register Prize do
   permit_params :name, :percentage, :image, :description, :identifier, :prize_type
+
+  controller do
+    def create
+      total_prizes_percentages = Prize.pluck(:percentage).sum
+      if total_prizes_percentages + params[:prize][:percentage].to_i > 100
+        render json: 'Error! The total percentage is greater than 100%!'
+      else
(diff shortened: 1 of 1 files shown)


app/admin/prizes.rb
```diff
@@ -2,4 +2,33 @@
 
 ActiveAdmin.register Prize do
   permit_params :name, :percentage, :image, :description, :identifier, :prize_type
+
+  controller do
+    def create
+      total_prizes_percentages = Prize.pluck(:percentage).sum
+      if total_prizes_percentages > 100 || total_prizes_percentages + params[:prize][:percentage].to_i > 100
```
zediogoviana, line 9: I think this first validation `total_prizes_percentages > 100` is redundant since you're already ensuring that when saving (create and update) the total sum is not over 100, so it seems that this case will never happen, because if `total_prizes_percentages > 100` then this plus the new value will for sure be above 100, right? (same thing for the update)

RicardoVaz1: Yes, I'll remove it

RicardoVaz1: Done

zediogoviana (approved):

PR #142, opened 2022-08-18
Add terms and conditions customization through active admin

beasrodrigues24:
Why: 
* We want the admin to be able to customize the terms and conditions link, used in the privacy policy page

How:
* By creating a model and migration for the terms and conditions, containing a link
* By generating the active admin resource and removing the create possibility (we want the admin to just re-edit the same one)
* By fixing the front-end to fetch this link from the database

### DIFF
--- a/app/controllers/terms_and_conditions_controller.rb
+++ b/app/controllers/terms_and_conditions_controller.rb
@@ -3,6 +3,10 @@
 # Controller that manages terms and conditions
 class TermsAndConditionsController < ApiController
   def index
+    render json: TermsAndCondition.all
+  end
+
+  def show
     render json: TermsAndCondition.find(1).slice(:link)
   end
 end
--- a/app/controllers/terms_and_conditions_controller.rb
+++ b/app/controllers/terms_and_conditions_controller.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+# Controller that manages terms and conditions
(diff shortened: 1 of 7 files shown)


app/controllers/terms_and_conditions_controller.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Controller that manages terms and conditions
+class TermsAndConditionsController < ApiController
+  def index
+    render json: TermsAndCondition.find(1).slice(:link)
```
zediogoviana, line 6: Why is the index hardcoded to `find(1)`?

beasrodrigues24: So we want there to only be one of the Terms and Conditions. I didn't know how to do that limitation any other way :sweat_smile:

zediogoviana: Then why not use `show` or some custom naming? I think it'd be more explicit given that you'll only receive one object, and not "all".

beasrodrigues24: Okay, I'll change it to show :+1:

beasrodrigues24: Done!!

zediogoviana (approved):

PR #143, opened 2022-08-19
Stop creating new Lead entries with the same email

beasrodrigues24:
Why:
* Instead of creating new entries with the same email, update the existing one with the new informations.
* This won't change the prize attributed and won't allow them to play again if they already have
* Allows the user to edit information they might have gotten wrong or changed their mind about

How:
* By updating the controller to check if the email exists -> if it does, update the existing entry, otherwise create a new one

### DIFF
--- a/app/controllers/leads_controller.rb
+++ b/app/controllers/leads_controller.rb
@@ -7,8 +7,13 @@ class LeadsController < ApiController
   end
 
   def create
-    lead = Lead.create(leads_params)
-    render json: lead
+    lead = Lead.find_by(email: leads_params[:email])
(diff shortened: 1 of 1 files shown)


zediogoviana (approved):
I don't have enough context on what a `Lead` is, and how it can be created/updated, but the code looks good 👍

beasrodrigues24:
> I don't have enough context on what a `Lead` is, and how it can be created/updated, but the code looks good +1

A Lead is a User, but we decided on "Lead" terminology because this isn't really a user since it can't be used to login, etc.

PR #145, opened 2022-08-19
Fix delete Prizes on Active Admin

RicardoVaz1:
Why:
* When we try to delete a Prize that's assigned to a Lead it'll fail

How:
* Fixing the destroy funciton on Active Admin
* Refactor the Prize controller code to be small

### DIFF
--- a/app/admin/prizes.rb
+++ b/app/admin/prizes.rb
@@ -5,30 +5,28 @@ ActiveAdmin.register Prize do
 
   controller do
     def create
-      total_prizes_percentages = Prize.pluck(:percentage).sum
-      if total_prizes_percentages + params[:prize][:percentage].to_i > 100
+      if Prize.pluck(:percentage).sum + params[:prize][:percentage].to_i > 100
         render json: 'Error! The total percentage is greater than 100%!'
       else
-        Prize.create(prize_params)
-        redirect_to '/admin/prizes'
-      end
+        Prize.create(params.require(:prize).permit(:name, :percentage, :image, :prize_type, :description, :identifier))
+        redirect_to '/admin/prizes' end
     end
 
     def update
-      prize_info = Prize.find_by(id: params[:id])
-      total_prizes_percentages = Prize.pluck(:percentage).sum - prize_info.percentage
-
-      if total_prizes_percentages + params[:prize][:percentage].to_i > 100
+      if (Prize.pluck(:percentage).sum - Prize.find_by(id: params[:id]).percentage) +
+         params[:prize][:percentage].to_i > 100
         render json: 'Error! The total percentage is greater than 100%!'
       else
-        super
-      end
+        super end
     end
 
-    private
-
-    def prize_params
-      params.require(:prize).permit(:name, :percentage, :image, :prize_type, :description, :identifier)
+    def destroy
+      if Lead.find_by(prize_id: params[:id])
+        Prize.find_by(id: params[:id]).update(percentage: 0)
+        render json: 'You can\'t destroy this prize, because you\'ll destroy that prize on Leads records!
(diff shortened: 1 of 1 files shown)


app/admin/prizes.rb
```diff
@@ -5,30 +5,28 @@
...
-      total_prizes_percentages = Prize.pluck(:percentage).sum
-      if total_prizes_percentages + params[:prize][:percentage].to_i > 100
+      if Prize.pluck(:percentage).sum + params[:prize][:percentage].to_i > 100
         render json: 'Error! The total percentage is greater than 100%!'
       else
-        Prize.create(prize_params)
-        redirect_to '/admin/prizes'
-      end
+        Prize.create(params.require(:prize).permit(:name, :percentage, :image, :prize_type, :description, :identifier))
+        redirect_to '/admin/prizes' end
```
zediogoviana, line 12: This `end`, and the ones in lines 20, 29 are badly placed. They should be in the line below.

app/admin/prizes.rb
```diff
@@ -5,30 +5,28 @@
...
     end
 
     def update
-      prize_info = Prize.find_by(id: params[:id])
-      total_prizes_percentages = Prize.pluck(:percentage).sum - prize_info.percentage
-
-      if total_prizes_percentages + params[:prize][:percentage].to_i > 100
+      if (Prize.pluck(:percentage).sum - Prize.find_by(id: params[:id]).percentage) +
+         params[:prize][:percentage].to_i > 100
```
zediogoviana, line 17: To be honest I think what you had before was better... Not every piece of code needs to be optimized to be the smallest possible but to be the easiest to understand, and I think that before was easier in that sense.

app/admin/prizes.rb
RicardoVaz1: But if I change that I'll have this error `Block has too many lines. [27/25] (convention:Metrics/BlockLength)`. 
Can I fix it with this `# rubocop:disable Metrics/BlockLength`?

app/admin/prizes.rb
RicardoVaz1: Same reason, I optimize the code because I was having this error: `Block has too many lines. [35/25] (convention:Metrics/BlockLength)`. Can I fix it with this `# rubocop:disable Metrics/BlockLength`?

app/admin/prizes.rb
zediogoviana: Check if it works on both cases 😅

RicardoVaz1: Done

app/admin/prizes.rb
```diff
@@ -5,30 +5,24 @@
...
     def create
-      total_prizes_percentages = Prize.pluck(:percentage).sum
-      if total_prizes_percentages + params[:prize][:percentage].to_i > 100
-        render json: 'Error! The total percentage is greater than 100%!'
-      else
-        Prize.create(prize_params)
-        redirect_to '/admin/prizes'
-      end
+      total_prizes_percentages = Prize.pluck(:percentage).sum + params[:prize][:percentage].to_i
+      total_prizes_percentages > 100 ? (render json: 'Error! The total percentage is greater than 100%!') : super
```
zediogoviana, line 9: What about the 
```
  Prize.create(prize_params)
  redirect_to '/admin/prizes'
```

you had in the `else` case before? Is it the same thing as `super`?

RicardoVaz1: Yes, it is. The only difference is that instead of redirecting to the Prizes List is redirecting to the Prize Details. And this way I reduced the code size, so it doesn't complain about code size

zediogoviana (approved):

PR #148, opened 2022-08-22
fix spacing

jreis13:
Why:
* Spacing was not matching the design.

How:
* Adjusted spacing to match design

### DIFF
--- a/frontend/src/components/Forms/FormLayout/index.module.css
+++ b/frontend/src/components/Forms/FormLayout/index.module.css
@@ -1,4 +1,4 @@
 .headerWrapper {
-  margin-bottom: 48px;
+  margin-bottom: 16px;
   text-align: center;
 }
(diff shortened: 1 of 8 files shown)


frontend/src/components/Forms/FormLayout/index.module.css
```diff
@@ -1,4 +1,4 @@
 .headerWrapper {
-  margin-bottom: 48px;
+  margin-bottom: 16px;
```
jreis13, line 2: this is in relation to a new line of text will be added in another PR

joaojusto (commented):
So this is the reference: 
Screenshot 2022-08-22 at 11 19 52

This is the running version on this PR:
Screen Shot 2022-08-22 at 11 21 09

It seems that font sizes are completely off as well as the input sizes.

joaojusto (approved):

PR #150, opened 2022-08-22
Change Wheel to a static image

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5250019007

How:
* Changing the Wheel to a static image


joaojusto (changes requested):
The button is different from the reference design, maybe we should also export the button background as an SVG. The font size and weight looks completely off:

Reference: 
Screenshot 2022-08-22 at 13 47 53

Implemented on this PR:
Screenshot 2022-08-22 at 13 50 18

joaojusto (approved):
LGTM, I would center the whole wheel on the page before merging 👍

PR #151, opened 2022-08-22
Review and update copy throughought app

beasrodrigues24:
Why: 
* Task: https://3.basecamp.com/4319812/buckets/26581088/todos/5250036530

How:
* By refactoring the LayoutWithBackGround to allow variation between svg wheel position in the left or right
* By viewing figma and matching

Note: Alignment will be fixed in another PR


joaojusto (approved):
LGTM

PR #155, opened 2022-08-22
Remove console errors

beasrodrigues24:
Why:
* Console was sending a lot of error messages

How:
* By reading them and refactoring them

### DIFF
--- a/frontend/src/components/Forms/DetailsForm/index.js
+++ b/frontend/src/components/Forms/DetailsForm/index.js
@@ -9,26 +9,28 @@ const DetailsForm = ({ error, handleChange, name, email }) => {
       headerText="A little bit about you"
       subtitleText="Because you will collect your prize via email"
     >
-      <Input
-        id="name"
-        name="Name"
-        type="text"
-        placeholder="Type in your name"
-        value={name}
-        onChange={handleChange("name")}
-        error={error.hasError}
-        errorMessage={error.message}
-      />
-      <Input
-        id="email"
-        name="Email"
-        type="email"
-        placeholder="Type in your email"
-        value={email}
-        onChange={handleChange("email")}
-        error={error.hasError}
-        errorMessage={error.message}
-      />
+      <>
+        <Input
+          id="name"
+          name="Name"
+          type="text"
+          placeholder="Type in your name"
+          value={name}
+          onChange={handleChange("name")}
+          error={error.hasError}
+          errorMessage={error.message}
+        />
+        <Input
+          id="email"
+          name="Email"
+          type="email"
+          placeholder="Type in your email"
+          value={email}
+          onChange={handleChange("email")}
+          error={error.hasError}
+          errorMessage={error.message}
+        />
+      </>
     </FormLayout>
   );
 };
--- a/frontend/src/components/Forms/DetailsForm/index.js
+++ b/frontend/src/components/Forms/DetailsForm/index.js
@@ -9,28 +9,26 @@ const DetailsForm = ({ error, handleChange, name, email }) => {
       headerText="A little bit about you"
       subtitleText="Because you will collect your prize via email"
     >
-      <>
-        <Input
-          id="name"
-          name="Name"
-          type="text"
-          placeholder="Type in your name"
-          value={name}
-          onChange={handleChange("name")}
-          error={error.hasError}
-          errorMessage={error.message}
-        />
-        <Input
-          id="email"
-          name="Email"
-          type="email"
-          placeholder="Type in your email"
-          value={email}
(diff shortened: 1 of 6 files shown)


frontend/src/components/Typography/Body/index.js
```diff
@@ -6,7 +6,7 @@ const Body = ({ children }) => {
 };
 
 Body.propTypes = {
-  children: PropTypes.string.isRequired,
+  children: PropTypes.string,
```
mirhamasala, line 9: Is it not required?

frontend/src/pages/HomePage/index.js
```diff
@@ -16,9 +16,8 @@ const HomePage = () => {
           </div>
 
           <div className={styles.headingWrapper}>
-            <Heading2>
-              Register to play and<br></br> win a prize!
-            </Heading2>
+            <Heading2>Register to play and</Heading2>
+            <Heading2>win a prize!</Heading2>
```
mirhamasala, line 20: Hmm. Nice try. But nope. 🙂

frontend/src/components/Typography/Body/index.js
beasrodrigues24: So there is a case where we might pass something to the body that is null depending on another thing.

frontend/src/pages/HomePage/index.js
beasrodrigues24: I tried with {"\n"} with {"aa\naa"} and with <br></br>. The only one that worked was the last one but it says I'm passing an array to Heading2

frontend/src/components/Typography/Body/index.js
mirhamasala: You made the error go away, but you didn't actually fix the core issue. Which is that an `<Input>` component was used without providing a `name`. It's `undefined`.

beasrodrigues24: Okay, on it!

beasrodrigues24: Done!

frontend/src/pages/HomePage/index.js
beasrodrigues24: Tried something I found on google about setting white-space: pre-line; and then using \n. Is that good?

frontend/src/components/Typography/Body/index.js
beasrodrigues24: Now that I think about it, with Joao's PR this error is probably gone anyways :thinking:

frontend/src/components/Forms/DetailsForm/index.js
```diff
@@ -9,26 +9,28 @@ const DetailsForm = ({ error, handleChange, name, email }) => {
...
+          id="email"
+          name="Email"
+          type="email"
+          placeholder="Type in your email"
+          value={email}
+          onChange={handleChange("email")}
+          error={error.hasError}
+          errorMessage={error.message}
+        />
+      </>
```
joaojusto, line 33: Why do we need the fragment here? 🤔

mirhamasala: I think it's because the `FormLayout` expects `children` to be `children: PropTypes.element.isRequired`

mirhamasala: `FormLayout` use to accept only one child, so this changed recently I believed.

beasrodrigues24: The truth is we don't need errors anymore, but I was thinking of removing errors in another PR so it's not too confusing

beasrodrigues24: It's that!

joaojusto: What about changing the prop definition to the correct one?

beasrodrigues24: I'm very confused with this PR now :sweat:

mirhamasala: I think what @joaojusto is trying to say is that you should think about what you're expecting `FormLayout` to receive as `children` and adapt the `propTypes` according to that instead of the other way around.

frontend/src/pages/HomePage/index.js
mirhamasala: Is this going to be solved with the `max-width` styles in your other commit?

beasrodrigues24: It's not but it can be solved in that way too, so I'll change it!

mirhamasala (approved):

PR #158, opened 2022-08-23
Change email from string to text

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5248215596

How:
* Changing email body from string to text

### DIFF
--- a/db/migrate/20220718133934_create_emails.rb
+++ b/db/migrate/20220718133934_create_emails.rb
@@ -3,7 +3,7 @@ class CreateEmails < ActiveRecord::Migration[7.0]
     create_table :emails do |t|
       t.string :email_type
       t.string :subject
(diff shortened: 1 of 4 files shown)


db/migrate/20220718133934_create_emails.rb
```diff
@@ -3,7 +3,7 @@ def change
     create_table :emails do |t|
       t.string :email_type
       t.string :subject
-      t.string :body
+      t.text :body
```
zediogoviana, line 6: You can't change an existing migration. This won't change anything for people who already have the DB. You need to create a new migration where you alter the column type...

zediogoviana (changes requested):

zediogoviana (approved):

PR #162, opened 2022-08-23
Refactor headings

mirhamasala:


RicardoVaz1 (approved):

frontend/src/components/Forms/FormLayout/index.js
```diff
@@ -1,17 +1,16 @@
...
 const FormLayout = ({ children, headerText, subtitleText }) => {
   return (
     <div className={styles.root}>
       <div className={styles.headerWrapper}>
-        <Heading1>{headerText}</Heading1>
+        <Heading>{headerText}</Heading>
       </div>
       <div className={styles.subtitleWrapper}>
-        <Heading4>{subtitleText}</Heading4>
+        <Heading level={4}>{subtitleText}</Heading>
```
mirhamasala, line 13: So perhaps now you can see why @joaojusto suggested talking to Hannah.

You know have a level 4 heading right after a level 1 heading. SEO wise, it's odd to be skipping heading levels like this.

The problem is that we're using level tags tied to stylistic purposes. We're treating `h4` as if it says something about how it should look. While, there's nothing stopping us from having proper level order, but, perhaps, tiny `h1` styles and much bigger `h4` styles.

Not saying you have to do anything about it now. But just want you to think about it! 🙂

mirhamasala: Instead of having just a `Heading` component, we could split it. Have a `MainHeading` component which always return an `h1` and the accompanying styles. And a `SubHeading` component, which would have the levels 2, 3, 4, 5, 6 but the classes wouldn't be tied to those numbers. Instead you might have something like, `xs`, `s`, `md`, `lg`.

mirhamasala: Again, no need to go changing it now. But I hope it shows how refactoring, progressively shoes you the patterns and errors in your code that are hard to spot when you still don't have enough code on the screen.

mirhamasala (commented):
@jreis13 Just a few comments to make you think further. Other than that, great job! LGTM. ☀️

PR #164, opened 2022-08-24
Change route names

jreis13:
Why:
* Route naming was inconsistent and too long.

How:
* Use simpler and more intuitive naming.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,7 +6,7 @@ Rails.application.routes.draw do
   get '/prizes', to: 'prizes#index'
   post '/prizes', to: 'prizes#create'
   get '/socials', to: 'socials#show'
-  post '/win-prize', to: 'prizes#win_prize'
+  post '/prize', to: 'prizes#win_prize'
   post '/leads', to: 'leads#create'
   get '/emails', to: 'emails#index'
   get '/terms', to: 'terms_and_conditions#show'
--- a/frontend/src/components/Wheel/index.js
+++ b/frontend/src/components/Wheel/index.js
@@ -15,7 +15,7 @@ const Wheel = ({ email }) => {
   async function startRotate() {
     setIsRotating((prev) => !prev);
 
-    await fetch(`${process.env.REACT_APP_BACKEND_PATH}/win-prize`, {
+    await fetch(`${process.env.REACT_APP_BACKEND_PATH}/prize`, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify({ email: email }),
@@ -33,7 +33,7 @@ const Wheel = ({ email }) => {
       .catch((error) => alert("Error during POST!", error));
 
     setTimeout(() => {
-      navigator("/win-prize");
+      navigator("/prize");
     }, 5000);
   }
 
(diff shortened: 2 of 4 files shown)


frontend/src/components/Wheel/index.js
```diff
@@ -15,7 +15,7 @@ const Wheel = ({ email }) => {
   async function startRotate() {
     setIsRotating((prev) => !prev);
 
-    await fetch(`${process.env.REACT_APP_BACKEND_PATH}/win-prize`, {
+    await fetch(`${process.env.REACT_APP_BACKEND_PATH}/prize`, {
```
mirhamasala, line 18: This needs to exist in the backend.

mirhamasala: In your Rails server console:

```
Started POST "/prize" for ::1 at 2022-08-24 11:18:56 +0200
  
ActionController::RoutingError (No route matches [POST] "/prize"):
```

mirhamasala: And in your browser console.

Screenshot 2022-08-24 at 11 22 52

mirhamasala: You can either also change the backend route to `/prize` or you can keep the backend route as `/win-prize`

config/routes.rb
```diff
@@ -6,7 +6,7 @@
   get '/prizes', to: 'prizes#index'
   post '/prizes', to: 'prizes#create'
   get '/socials', to: 'socials#show'
-  post '/win-prize', to: 'prizes#win_prize'
+  post '/prize', to: 'prizes#win_prize'
```
mirhamasala, line 9: @zediogoviana Any restful / Rails conventions violated here?

frontend/src/components/Wheel/index.js
jreis13: I think I fixed with [this](https://github.com/Subvisual-Academy/subvisuwheel/pull/164/commits/c60ae90a8a5e6072899466c929c849a4fff79462) just before you commented or am I wrong. Got confused 😄

config/routes.rb
zediogoviana: What does this `win_prize` do? Either way is not restful because it should only be only `CRUD` operations... I believe this would have to be namespace into `leads` and do something like `post '/leads/:lead/prize', to: lead_prizes#create` to be restful, because I'd be creating a `prize` for that specific `lead`. 
But this would require more changes than just updating the route.

zediogoviana: Link to [Rails docs](https://guides.rubyonrails.org/routing.html#nested-resources)

frontend/src/components/Wheel/index.js
mirhamasala: You're right!

config/routes.rb
mirhamasala: Got it! Will leave this change for now as is.

If you are curious about making it restful, perhaps you can start a thread in Slack. @beasrodrigues24 @RicardoVaz1 @jreis13 (This would be backend refactoring)

mirhamasala: Thank you @zediogoviana 💙

mirhamasala (approved):

PR #166, opened 2022-08-24
Refactor form

mirhamasala:


beasrodrigues24:
This does look better, having label and the input together!

beasrodrigues24:
Is the FormLabel a component as well to isolate better its styling or what is exacly the reason? Just curiosity

beasrodrigues24 (approved):

mirhamasala:
> Is the FormLabel a component as well to isolate better its styling or what is exacly the reason? Just curiosity

Styles and functionality. In this case, I named it `FormLabel` but you can imagine that it could also be a generic `Label` that's not only used in Forms.

Also, this way, you can have layout components like `FormField` and others, that can position the `FormLabel` with the `Input` differently. Without you having to keep adding classes to do so to `Input`.

Lastly, it really is a component of its own, isn't it? A label is not an input.

PR #167, opened 2022-08-24
Fix spacing in form

jreis13:
Why:
* We were using a negative margin in the caption because we were applying the `margin-bottom` on the `Input` component.

How:
* Moved the `margin-bottom` to the `FormField` component and removed the negative margin from the `FormHint` component.

### DIFF
--- a/frontend/src/components/FormField/index.module.css
+++ b/frontend/src/components/FormField/index.module.css
@@ -1,7 +1,6 @@
 .root {
   display: grid;
-  grid-template-rows: repeat(2, auto);
+  grid-template-rows: auto;
   grid-gap: 12px;
   width: 100%;
-  margin-bottom: 32px;
 }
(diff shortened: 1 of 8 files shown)


frontend/src/components/FormField/index.module.css
```diff
@@ -3,4 +3,5 @@
   grid-template-rows: repeat(2, auto);
   grid-gap: 12px;
   width: 100%;
+  margin-bottom: 32px;
```
mirhamasala, line 6: Hmm nope. This is a self-contained component. You should be able to drop it anywhere. Meaning, that you probably have to look at the parent. Or create a `FormFieldGroup` / `FormFieldLayout` component.

frontend/src/components/FormField/index.module.css
```diff
@@ -3,4 +3,5 @@
   grid-template-rows: repeat(2, auto);
```
mirhamasala, line 3: This could be just `auto` now with the hint.

mirhamasala (approved):
Excellent! ✨🚀

PR #169, opened 2022-08-24
Replace subtitleText component

jreis13:
Why:
* We were using an `h4` and the design is using a `<Body>` component.

How:
* Replaced the `<Heading>` component for the `<Body>` component.

### DIFF
--- a/frontend/src/components/Forms/FormLayout/index.js
+++ b/frontend/src/components/Forms/FormLayout/index.js
@@ -1,17 +1,17 @@
 import { PropTypes } from "prop-types";
 import Heading from "components/Typography/Heading";
-import Body from "components/Typography/Body";
+import Subtitle from "components/Typography/Subtitle";
 
 import styles from "./index.module.css";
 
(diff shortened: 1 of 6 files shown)


frontend/src/components/Forms/FormLayout/index.js
```diff
@@ -10,7 +11,7 @@ const FormLayout = ({ children, headerText, subtitleText }) => {
         <Heading>{headerText}</Heading>
       </div>
       <div className={styles.subtitleWrapper}>
-        <Heading level={4}>{subtitleText}</Heading>
+        <Body>{subtitleText}</Body>
       </div>
```
mirhamasala, line 15: Is it still a `SubTitle` then? Or should the `Body` component be a `SubTitle` component?

mirhamasala:
I don't like it much. 🙃 I think this should be an `h2` (subTitle) with the styles from `<Body>`. Semantically that's what makes sense to me.

mirhamasala (approved):
LGTM! ✨

PR #171, opened 2022-08-24
Clean up

mirhamasala:
Why:

- Some code could use a clean up

How:

- Rename `Input` to `FormInput` to be consistent with other Form components
- Use color variables when possible
- Remove `constants` file and unused config
- Use `LOGOS` directly in social page
- Use self-closing tags


mirhamasala:
Please have a look, @beasrodrigues24 @RicardoVaz1 @jreis13. The clean-up PRs are for you just to learn/follow along what could be done still. It's a mixed bag. Nothing crazy.

beasrodrigues24 (approved):
It does look great. And this made me realize that yes made no sense for the logos to be in that file when they were only used on the social page :sweat_smile: We should've noticed that!

jreis13 (approved):

PR #172, opened 2022-08-24
Adapt prize attribution

beasrodrigues24:
Why: 
* Prizes should be attributed in a "Update" inside of the leads controller
* This attribution should be based on ID 

How: 
* By moving the prizes attribution to the leads controller 
* By going from the name win_prize to the name update to respect CRUD
* By adapting the config/routes.rb file to a patch request to leads/
* By fixing ValidateUser to have the right name (copy paste mistake)
* By changing localStorage.setItem("userData") to "userID" because it went from being an email to being an ID
* By making the post request a patch request

### DIFF
--- a/app/controllers/leads_controller.rb
+++ b/app/controllers/leads_controller.rb
@@ -13,10 +13,68 @@ class LeadsController < ApiController
     else
       lead = Lead.create(leads_params)
     end
-    render json: lead
+    render json: lead.slice(:id)
   end
+
+  def update
+    lead = Lead.find(params[:id])
+
+    if lead.prize.nil?
+      won_prize = sort_prize
+      name_lead = assign_prize_to_lead(won_prize)
+      send_email(name_lead, won_prize)
+    end
+    render json: lead.prize.slice(:name, :prize_type)
+  end
+
+  private
 
   def leads_params
     params.require(:lead).permit(:name, :email, [interests: []], :job_consent)
+  end
+
+  def sort_prize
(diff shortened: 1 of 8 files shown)


app/controllers/leads_controller.rb
```diff
@@ -13,10 +13,68 @@ def create
...
     params.require(:lead).permit(:name, :email, [interests: []], :job_consent)
   end
+
+  def sort_prize
+    prizes_list = Prize.all.each_with_object({}) do |prize, acc|
+      acc[prize.name] = prize.percentage
+    end
+
+    Pickup.new(prizes_list).pick
+  end
```
zediogoviana, line 42: If you have the time, it might be interesting to move this and the code below to the `Prize` and `Lead` models... And then you could do something like `Prize.random_select` or `lead.assign_prize(prize)`, etc. Basically a more OOP oriented programming (as is the case with Ruby).

zediogoviana (approved):

app/controllers/leads_controller.rb
beasrodrigues24: I think I might've some time in the afternoon for that so I won't merge for now to see if I can still do that. Is that something similiar to scopes? What should I read about to help me do this?

zediogoviana: You can still merge this one, and open a new one later

beasrodrigues24: okay!

PR #178, opened 2022-08-25
Fix wheel not spinning in some devices

RicardoVaz1:
Why:
* https://3.basecamp.com/4319812/buckets/26581088/todos/5262079540

How:
* Adding webkit in the css animation

### DIFF
--- a/frontend/src/components/Forms/DetailsForm/index.js
+++ b/frontend/src/components/Forms/DetailsForm/index.js
@@ -14,7 +14,7 @@ const DetailsForm = ({ error, handleChange, name, email }) => {
     >
       <FormFieldLayout>
         <FormField
-          label={<FormLabel id="name">Name</FormLabel>}
+          label={<FormLabel id="name">Name*</FormLabel>}
           input={
             <FormInput
               id="name"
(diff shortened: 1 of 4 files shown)


frontend/src/components/Forms/DetailsForm/index.js
```diff
@@ -14,7 +14,7 @@ const DetailsForm = ({ error, handleChange, name, email }) => {
     >
       <FormFieldLayout>
         <FormField
-          label={<FormLabel id="name">Name</FormLabel>}
+          label={<FormLabel id="name">Name*</FormLabel>}
```
mirhamasala, line 17: Shall we do this in a different PR? Would also be nice to render these based on the `required` attribute. Google a bit around how those are styles in React/CSS.


https://www.w3schools.com/tags/att_input_required.asp

mirhamasala: This would also require talking to Hannah. So no sure if we actually want to do this?

mirhamasala (approved):
Just remove the required stuff, and this is good to go! 🚀

PR #180, opened 2022-08-25
Improve prize attribution

beasrodrigues24:
Why:
* It could be arrange to be instance and class methods

How:
* By creating a class method named random_select that returns a prize
* By creating an instance method named assign_prize that awards a lead a prize

### DIFF
--- a/app/models/lead.rb
+++ b/app/models/lead.rb
@@ -4,8 +4,4 @@
 class Lead < ApplicationRecord
   validates :name, :email, presence: true
   belongs_to :prize, optional: true
-
-  def assign_prize(prize_won)
-    update!(prize: prize_won)
-  end
 end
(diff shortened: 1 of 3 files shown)


beasrodrigues24:
My question is: the sending email methods can also be arranged elsewhere? What else could be improved here?

app/models/lead.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
+# Class that holds lead information
 class Lead < ApplicationRecord
   validates :name, :email, presence: true
   belongs_to :prize, optional: true
+
+  def assign_prize(prize_won)
```
zamith, line 8: If it's this one line, why not call the update directly?

app/models/lead.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
+# Class that holds lead information
```
zamith, line 3: Do you think this is an important comment?

beasrodrigues24: I was forced by rubocop

